PR #4, opened 2020-10-14
Stub server fix for single threading

technige:
Also replaces `<EXIT>` with `<BREAK>` for clarity on functionality.


2hdddg:
We have removed all script files and moved the scripts into the python code right where the tests are, that is what is causing the conflict. Easiest is probably to revert changes to .script files and rebase this branch and make the same changes again in the corresponding python files.

PR #5, opened 2020-10-19
Fix container removal between tests on Mac

bigmontz:
The neo4j server container was not being removed between tests and it was causing name conflicts.


main.py
```diff
@@ -275,6 +275,8 @@ def ensure_driver_image(root_path, branch_name, driver_name):
 
         subprocess.run([
             "docker", "stop", neo4jServerHostname])
+        subprocess.run([
```
2hdddg, line 278: We can remove the docker stop and instead do "docker rm -f -v ..." Then we don't need to wait for the timeout and we don't care about the data in the container anyway

bigmontz: Fixed.

PR #6, opened 2020-10-23
Updating routing tests to support new Route Message 

bigmontz:
The scripts which originaly was from the Routing were moved to RoutingV4 and Routing was updated with new scripts using the Route message. The new tests for the version 4.3 is being skiped for any language but java since it's the only implementing the new message.

The support for the new message was added in scripting.Bolt4x3Script and register to be used only in the version 4.3

### DIFF
--- a/tests/stub/routing.py
+++ b/tests/stub/routing.py
@@ -18,6 +18,8 @@ def get_extra_hello_props():
 # to handle variations.
 class Routing(unittest.TestCase):
     def setUp(self):
+        if type(self) is Routing and get_driver_name() not in ['java']:
+            self.skipTest('Routing message not implemented in driver')
         self._backend = new_backend()
         self._routingServer = StubServer(9001)
         self._readServer = StubServer(9002)
@@ -25,7 +27,7 @@ class Routing(unittest.TestCase):
         self._uri = "neo4j://%s?region=china&policy=my_policy" % self._routingServer.address
         self._auth = AuthorizationToken(scheme="basic", principal="p", credentials="c")
         self._userAgent = "007"
-
+        
     def tearDown(self):
         self._backend.close()
         self._routingServer.reset()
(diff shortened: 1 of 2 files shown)


main.py
```diff
@@ -218,9 +218,9 @@ def ensure_driver_image(root_path, branch_name, driver_name):
     os.makedirs(neo4jArtifactsPath)
 
     neo4jServers = [
-        { "version": "3.5", "edition": "enterprise", "cluster": False, "suite": "3.5", "scheme": "bolt" },
         { "version": "4.0", "edition": "community", "cluster": False, "suite": "4.0", "scheme": "neo4j" },
         { "version": "4.1", "edition": "enterprise", "cluster": False, "suite": "4.1", "scheme": "neo4j" },
+        { "version": "4.1", "edition": "enterprise", "cluster": False, "suite": "4.3", "scheme": "neo4j" }
```
2hdddg: We want this soon but for now it will just take extra time and resources from the build environment. We should add  this when there are docker images available for 4.3 (not  necessariliy via dockerhub)

tests/stub/routing.py
```diff
@@ -18,14 +18,16 @@ def get_extra_hello_props():
 # to handle variations.
 class Routing(unittest.TestCase):
     def setUp(self):
+        if type(self) is Routing and get_driver_name() not in ['java']:
+            self.skipTest('Routing message not implemented in backend')
```
2hdddg: Nitpick, not implemented in driver

tests/stub/routing.py
```diff
@@ -185,13 +187,48 @@ def test_tx_run_write(self):
...
+        v = {
+            "#VERSION#": 4.1,
+            "#HOST#": host,
+            "#DBPARAM#": dbparam,
+            "#EXTRA_HELLO_PROPS#": extra,
+            "#ROUTINGMODE#": "",
+            "#ROUTINGCTX#": '{"address": "' + host + ':9001", "region": "china", "policy": "my_policy"}',
+        }
+        if get_driver_name() in ['go']:
+            v["#ROUTINGMODE#"] = '"mode": "r"'
```
2hdddg: I thought I had fixed that in Go driver, might have been removed on 4.2

2hdddg (approved):

technige (approved):

PR #11, opened 2020-11-05
Support Routing Stub Tests for Javascript Driver

bigmontz:
Enable the tests and make the necessary adjustments on it.

### DIFF
--- a/tests/stub/routing.py
+++ b/tests/stub/routing.py
@@ -7,7 +7,11 @@ import nutkit.protocol as types
 
 
 def get_extra_hello_props():
-    return ', "realm": ""' if get_driver_name() in ["java"] else ""
+    if get_driver_name() in ["java"]:
+        return ', "realm": ""'
+    elif get_driver_name() in ["javascript"]:
+        return ', "realm": "", "ticket": ""'
+    return ""
 
 # This should be the latest/current version of the protocol.
 # Older protocol that needs to be tested inherits from this and override
@@ -34,7 +38,7 @@ class Routing(unittest.TestCase):
         !: AUTO RESET
         !: AUTO GOODBYE
 
-        C: HELLO {"scheme": "basic", "credentials": "c", "principal": "p", "user_agent": "007", "routing": #ROUTINGCTX# #EXTRA_HELLO_PROPS#}
+        C: HELLO {"scheme": "basic", "credentials": "c", "principal": "p", "user_agent": "007", "routing": #HELLO_ROUTINGCTX# #EXTRA_HELLO_PROPS#}
         S: SUCCESS {"server": "Neo4j/4.0.0", "connection_id": "bolt-123456789"}
         C: RUN "CALL dbms.routing.getRoutingTable($context, $#DBPARAM#)" {"context": #ROUTINGCTX#, "#DBPARAM#": "adb"} {"mode": "r", "db": "system"}
         C: PULL {"n": -1}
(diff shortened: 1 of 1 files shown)


tests/stub/routing.py
```diff
@@ -125,7 +135,7 @@ def get_db(self):
     # Checks that routing is used to connect to correct server and that parameters for
     # session run is passed on to the target server (not the router).
     def test_session_run_read(self):
-        if get_driver_name() not in ['go', 'dotnet', 'java']:
+        if get_driver_name() not in ['go', 'dotnet', 'java', 'javascript']:
```
2hdddg: We can remove this one

tests/stub/routing.py
```diff
@@ -136,11 +146,11 @@ def test_session_run_read(self):
         session.close()
         driver.close()
         self._routingServer.done()
-        self._readServer.done()
+        self._readServer.done() 
 
     # Same test as for session.run but for transaction run.
     def test_tx_run_read(self):
-        if get_driver_name() not in ['go', 'dotnet', 'java']:
+        if get_driver_name() not in ['go', 'dotnet', 'java', 'javascript']:
```
2hdddg: Remove

tests/stub/routing.py
```diff
@@ -157,7 +167,7 @@ def test_tx_run_read(self):
 
     # Checks that write server is used
     def test_session_run_write(self):
-        if get_driver_name() not in ['go', 'dotnet', 'java']:
+        if get_driver_name() not in ['go', 'dotnet', 'java', 'javascript']:
```
2hdddg: Remove

tests/stub/routing.py
```diff
@@ -172,7 +182,7 @@ def test_session_run_write(self):
 
     # Checks that write server is used
     def test_tx_run_write(self):
-        if get_driver_name() not in ['go', 'dotnet', 'java']:
+        if get_driver_name() not in ['go', 'dotnet', 'java', 'javascript']:
```
2hdddg: Remove

tests/stub/routing.py
```diff
@@ -115,8 +119,14 @@ def get_vars(self):
...
             "#EXTRA_HELLO_PROPS#": get_extra_hello_props(),
         }
-        if get_driver_name() in ['dotnet', 'java']:
+        v["#HELLO_ROUTINGCTX#"] = v["#ROUTINGCTX#"]
+
+        if get_driver_name() in ['dotnet', 'java', 'javascript']:
             v["#DBPARAM#"] = "database"
+
+        if get_driver_name() in ['javascript']:
+            v["#HELLO_ROUTINGCTX#"] = '{"region": "china", "policy": "my_policy"}'
```
2hdddg, line 128: Not related to the PR, but why is this only needed on js driver?

tests/stub/routing.py
```diff
@@ -313,14 +329,15 @@ def get_vars(self):
...
+            "#ROUTING#": '", routing": null' if get_driver_name() not in ['javascript'] else ''
         }
 
 
     # Checks that routing is disabled when URI is bolt, no routing in HELLO and
     # no call to retrieve routing table. From bolt >= 4.1 the routing context is
     # used to disable/enable server side routing.
     def test_session_run_read(self):
-        if get_driver_name() not in ['go', 'dotnet', 'java']:
+        if get_driver_name() not in ['go', 'dotnet', 'java', 'javascript']:
```
2hdddg: Remove

tests/stub/routing.py
```diff
@@ -275,10 +285,16 @@ def get_vars(self):
...
         }
+        
         if get_driver_name() in ['go']:
             v["#ROUTINGMODE#"] = '"mode": "r"'
+        
         if get_driver_name() in ['java']:
             v["#EXTR_HELLO_ROUTING_PROPS#"] = ', "routing": ' + v['#ROUTINGCTX#'] 
+        
+        if get_driver_name() in ['javascript']:
+            v["#ROUTINGCTX#"] = '{"region": "china", "policy": "my_policy"}'
```
2hdddg, line 284: Interesting, why is this only in js driver?

tests/stub/routing.py
bigmontz: python is not supported by testkit?

tests/stub/routing.py
bigmontz: It's a strange behaviour. 

On the 4.x, it sends the context with the `address` when it call the procedure,  but it sends the routing context without the `address` to the HELLO message. 

On 3, it doesn't send the `address` to the procedure and the java driver is the only one to send routing context on the HELLO. 

I think the lack of address on the routing message for the 3.x could be a bug/miss-behaviour. But I don't know the impact of this.

tests/stub/routing.py
2hdddg: No backend yet...

PR #19, opened 2020-11-20
Verify bolt protocol versions that driver supports

2hdddg:


boltstub/scripting.py
```diff
@@ -81,24 +82,35 @@ def auto_match(self, tag):
     def on_auto_match(self, request):
         raise NotImplementedError
 
+    def _get_version(self, request, n):
+        minorindex = 4 + (n * 4) + 2
```
bigmontz: I could be a good idea put this magic numbers in some constants to make it easier to understand what's happening.

bigmontz (approved):
🗡️

technige (approved):
All looks good to me :+1:

PR #20, opened 2020-11-23
Bring java 4.2 up to speed with current tests

gjmwoods:


tests/stub/retry.py
```diff
@@ -138,6 +138,7 @@ def test_disconnect_on_commit(self):
         # Should NOT retry when connection is lost on unconfirmed commit.
         # The rule could be relaxed on read transactions therefore we test on writeTransaction.
         # An error should be raised to indicate the failure
+        # ex JAVA - Keeps retrying on commit despite connection being dropped.
```
2hdddg: Perhaps add a skip with the text in the comment, like
if self._driverName in ["java"]:
    self.skipTest("Keeps retrying...

tests/neo4j/datatypes.py
```diff
@@ -123,7 +113,7 @@ def testShouldEchoNode(self):
 
     # Work in progress
     def testShouldEchoVeryLongMap(self):
-        if get_driver_name() not in ['dotnet', 'go', 'javascript']:
+        if get_driver_name() not in ['dotnet', 'go', 'javascript', 'java']:
```
2hdddg: For all of those "not in [four languages]" we could remove the check completely. That will perhaps be a pain when adding support for Python but it is a bit repetetive with all of those..

boltstub/scripting.py
```diff
@@ -460,8 +460,25 @@ def default_action(cls, actor, line=None):
 
     def match(self, message):
         tag = self.script.tag("C", self.tag_name)
-        return tag == message.tag and tuple(self.fields) == tuple(message.fields)
+        return tag == message.tag and self.compare(tuple(self.fields), tuple(message.fields))
 
+    def compare(self, script_fields, message_fields):
+        for i in range(len(script_fields)):
```
2hdddg: What about case where there are more entries in message_fields? In most cases the ordering of entries would cause apples to be compared to bananas but if the extra entry happens to be last it would pass by unnoticed

gjmwoods: added a check on tuple length

PR #22, opened 2020-11-25
Enable browser, stress and integration tests for javascript

bigmontz:
The ipv6 support flag is necessary to skip the ipv6 tests when it is not supported.

### DIFF
--- a/driver/javascript/integration.py
+++ b/driver/javascript/integration.py
@@ -1,4 +1,11 @@
+import subprocess, os
+
+def run(args):
+    subprocess.run(
+        args, universal_newlines=True, stderr=subprocess.STDOUT, check=True)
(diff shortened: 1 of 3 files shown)


main.py
```diff
@@ -170,11 +170,13 @@ def main(thisPath, driverName, testkitBranch, driverRepo):
     password = "pass"
     neo4jServerHostname = "neo4jserver"
     port = 7687
+    ipv6Enabled = False
```
2hdddg: I don't think that we will test ipv6 at this level, too costly to have ipv6 as a test matrix variable. If you need this env variable I suggest you move it in to the driver/javascript/* file

2hdddg (approved):

PR #25, opened 2020-12-08
Add support to the extended handshake

bigmontz:
* [x] Treat 4.2 as a separated protocol, this way given an error when it is not sent in the handshake
* [x] Fix the error message build (to not get error when it is missing script and other properties)
* [x] Create a test to check if the driver implements the version 4.2
* [x] Accept/implement the new extend handshake format
* [x] Make the test fail properly when the version is not there

### DIFF
--- a/boltstub/scripting.py
+++ b/boltstub/scripting.py
@@ -53,8 +53,10 @@ class BoltScript:
             return super().__new__(Bolt3Script)
         elif version in {(4,), (4, 0)}:
             return super().__new__(Bolt4x0Script)
-        elif version in {(4, 1), (4, 2)}:
+        elif version in {(4, 1)}:
             return super().__new__(Bolt4x1Script)
+        elif version in {(4, 2)}:
+            return super().__new__(Bolt4x2Script)
         elif version in {(4, 3)}:
             return super().__new__(Bolt4x3Script)
         else:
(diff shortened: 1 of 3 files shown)


boltstub/scripting.py
```diff
@@ -84,14 +86,22 @@ def auto_match(self, tag):
...
         major_offset = 3  # Major version offset relative to start of version
         versionindex = versions_offset + (n * version_size)
         return (request[versionindex + major_offset],
-                request[versionindex + minor_offset])
+                request[versionindex + minor_offset],
+                request[versionindex + range_offset])
+
+    def _get_versions(self, request): 
+        for n in range(0, 4):
+            (major, max_minor, minor_offset) = self._get_version_range_defintion(request, n)
```
2hdddg: Can't remember what the spec said... is minor_offset number of prior versions to support or the absolute min version? Judging by max_minor - minor_offset I suppose it is the number of back versions. Is it really right with an extra -1 in range?

bigmontz: minor_offset is the number of previous minor versions ([see](https://docs.google.com/document/d/1BQraZtw0_o_tEezCJE8-5RsL6d03Lg3bXTezwwIdUy4/edit)  

The range is open to the right, so a range(1, 2) will be [1], (1, 3) will be [1, 2], (3, 2, -1) will be [3]

2hdddg: Looks fine! Maybe a better name for minor_offset, it makes sense when offsetting in request but not so much here.
I like the look of this in on_handshake!

bigmontz: I've changed the name to minor_difference, since it's the name used in the spec.

2hdddg (approved):

PR #26, opened 2020-12-09
Preparations for running tests against clusters

2hdddg:


neo4j.py
```diff
@@ -0,0 +1,109 @@
+import docker
+import time
+import atexit
+from os.path import join, abspath
+
+
+class Server:
```
bigmontz, line 7: Maybe it could be called `Standalone` or `StandaloneServer` to be more precise. It's not obvious to know the role of this class without take a more careful look at the implementation.

bigmontz (approved):
🍵

neo4j.py
2hdddg: Good suggestion!
I have continued working on this so I fix it in later commit and merge this for now...

PR #30, opened 2020-12-16
Swapping .Net over to use .Net 5.0. Added command line arguments to enable selection of test groups and servers to run.

AndyHeap-NeoTech:


bigmontz:
@AndyHeap-NeoTech, could you provide some common usage examples on the readme.md?

2hdddg (commented):
We should default to everything being enabled, now it looks like everything is default off. So we can call them SKIP_xxx instead. 
To avoid the need to change in too many places when adding removing server configs we can have a SKIP_CONFIGS (better name needed) that is a , separated string that we check for existance in before running.
Alternative to command line param (which I usually prefer) is to use environment variables for controlling what to skip. The upside of this is that it would work with custom runs in teamcity without any change...

AndyHeap-NeoTech:
If no parameters are supplied all the flags are set to true. Behaviour should not change from the current unless command line arguments are supplied.

I'll put some examples in the readme as suggested.

AndyHeap-NeoTech:
I've pushed a refactor of the configuration selection (previously known as servers). The configs and their selection are in one location in the code now and the command line selects which to run by specifying the name field. Should be much simpler to maintain now. Thanks for the suggestion Peter.

main.py
```diff
@@ -15,10 +15,128 @@
...
+                   "suite": "4.0",
+                   "scheme": "neo4j"},
+                  {"name": "4.1-enterprise",
+                   "image": "neo4j:4.1-enterprise",
+                   "version": "4.1",
+                   "edition": "enterprise",
+                   "cluster": False,
+                   "suite": "4.1",
+                   "scheme": "neo4j"},
+                  {"name": "4.2-tc-enterprise",
```
2hdddg: This one should only be added when teamcity is enabled. Otherwise it will fail strangely when running on host with default setup without having configured teamcity

main.py
```diff
@@ -15,10 +15,128 @@
...
+                   "image": "neo4j:4.2.2-enterprise",
+                   "version": "4.2",
+                   "edition": "enterprise",
+                   "cluster": False,
+                   "suite": "4.2",
+                   "scheme": "neo4j",
+                   "download": teamcity.DockerImage("neo4j-enterprise-4.2.2-docker-loadable.tar")},
+                  {"name": "4.3-tc-enterprise",
```
2hdddg: Same as above

main.py
```diff
@@ -15,10 +15,128 @@
...
+
+def convert_to_str(input_seq, separator):
+    # Join all the strings in list
+    final_str = separator.join(input_seq)
+    return final_str
+
+
+def construct_configuration_list(requested_list):
+    # if no configs were requested we will default to adding them all
+    if not requested_list:
```
2hdddg, line 111: requested_list = config
return

main.py
```diff
@@ -15,10 +15,128 @@
 import docker
 import teamcity
 import neo4j
-
+import argparse
 
 networks = ["the-bridge"]
 
+TestFlags = {
```
2hdddg: Uppercase naming indicates class name... Feels strange to read ;-)
should be test_flags or perhaps testFlags (nice mix of these already...)

main.py
```diff
@@ -15,10 +15,128 @@
...
+    "UNIT_TESTS": True,
+    "INTEGRATION_TESTS": True,
+    "STUB_TESTS": True,
+    "STRESS_TESTS": True,
+    "TLS_TESTS": True
+}
+
+ConfigurationsToRun = []
+
+Configurations = [{"name": "4.2-cluster",
```
2hdddg: configurations

main.py
```diff
@@ -15,10 +15,128 @@
...
+    # if no configs were requested we will default to adding them all
+    if not requested_list:
+        requested_list = []
+        for config in Configurations:
+            requested_list.append(config["name"])
+
+    # Now try to find the requested configs and check they are available with current teamcity status
+    for config in Configurations:
+        if config["name"] in requested_list:
+            if "download" in config:
```
2hdddg: No need for this is if you check in_teamcity when you build the list, then the command line help will be right as well

main.py
AndyHeap-NeoTech: Will correct the formatting.

main.py
AndyHeap-NeoTech: This is added to the configurations_to_run list only when in_teamcity.  This initial declaration is a 'master' list of all configs that are then selected from in the construct_configuration_list function.

main.py
AndyHeap-NeoTech: This is the building of the list. The resulting configurations_to_run list is what is used later on.

2hdddg: Yes but if you build the initial list based on teamcity then invoking command line help will only list the available configs in your environment and this code can be removed. Win win

PR #33, opened 2021-01-14
Enabling test_client_exception_rolls_back_change

AndyHeap-NeoTech:


2hdddg:
This commit history needs to be cleaned up a bit. 25 commits to change one line in one file feels like 24 commits too many

PR #42, opened 2021-01-25
javascript: Remove support of TxRun.test_nested since the order of record request is not guaranteed

bigmontz:


2hdddg (approved):
Nice cleanup of the testcode after adding the skip!

PR #43, opened 2021-01-26
Test retry in cluster environments

bigmontz:
It's need to check how the retries will influence the routing tables refresh to avoid keep retrying to same failed instance. The `TestRetry` was kept with the common retry cases in non cluster environments and the typical cluster retry scenario were move to `TestRetryClustering` along with some basic retry scenarios.


2hdddg (approved):
Great!

PR #47, opened 2021-01-27
Forwarding parameters from run_all.py to main.py to make easier to run a custom set of tests

bigmontz:


2hdddg (approved):
Looks fine!
Are you sorting the drivers in order of lest probable failure now? :-)

Maybe you should enable some kind of pep8 checker in your editor, it is quite yellow (warnings) when i look at that file. There are pleny of other files that are bad as well so no worry...

bigmontz:
Yes, I am sorting this way and I was checking the java driver was compliant with the retry tests. 

I will configure the style checker and push it.

PR #52, opened 2021-01-28
Add some art to run all

2hdddg:


run_all.py
```diff
@@ -61,11 +69,24 @@ def run():
...
+                print(c*scale, end='')
+            print('')
+    print('')
+
 
 if __name__ == "__main__":
     (temp_path, driver_repo_path, branch) = setup_environment()
 
     for driver in drivers:
+        print_art(driver.get("art", ""), 2)
```
bigmontz: You're already checking if art exists, so you don't need to set a default for art.

bigmontz (approved):
🧑‍🎨 🎨

run_all.py
AndyHeap-NeoTech: I find the lines pleasing and the theme thought provoking. The palette is a little monochromatic for my taste though.

AndyHeap-NeoTech (approved):

run_all.py
2hdddg: I'm not sure it is worth the effort to add colours.. As far as I know logs in TeamCity will not render that anyway ;-)

PR #63, opened 2021-02-25
Fix boltstub client msg check

robsdedude:
Fixes incomplete check of drivers' messages sent to the stub server. A lot of improper behavior could've passed the tests this way.

Also contains an update to the ROUTE messages in the stub scripts as they expected wrong behavior of the drivers. It just never surfaced because of the bug mentioned above..


robsdedude:
This is superfluous as of #70

PR #70, opened 2021-03-11
Stub script extension

robsdedude:
Rewrote the stub server to support a row of new script directives:

Allow the driver to connect again to this server and play the same script.
E.g., to ask for a routing table again.
Note that still only one connection can be open at a time. The script has
to be finished or the client must disconnect before a new connection can
be made. E.g. to ask for a routing table again:
```
!: ALLOW RESTART
```

Allow multiple concurrent connections to the server. All connections will
start at the beginning of the script and have their own server state.
Note that this option implies `!: ALLOW RESTART` as allowing concurrent
connections but only allowing the script to be run once makes little sense:
```
!: ALLOW CONCURRENT
```

Alternative script paths:
```
{{
    THIS
----
    OR THAT
}}
```

Parallel script paths:
```
{{
    C: THIS_REQUEST
    S: THIS_RESPONSE
    C: THEN_THIS_OTHER_REQUEST
    S: THEN_THIS_OTHER_RESPONSE
    +++
    C: SOME_WHEN_THAT_REQUEST_TOO
    S: THAN_THAT_RESPONSE_TOO
}}
S: WHEN_BOTH_BRANCHES_DONE_THIS_RESPONSE
```

Any number of (incl. 0) repetitions:
```
{*
    C: RESET {}
    S: SUCCESS {}
*}
```

At least one repetition:
```
{+
    C: RESET {}
    S: SUCCESS {}
+}
```

Optional fields in dictionaries:
```
C: MSG {"[optional]": "value", "\[name_that_is_actually_surrounded_by_brackets\]": "value"}
```

Wildcard values:
```
C: MSG {"value_and_type_don't_matter": "*", "value_actually_is_*": "\*"}
```
or
```
C: MSG "*"
```
Unordered lists:
```
C: MSG {"key{}": [1, 2], "name_that_is_actually_ends_with_curly_brackets\{\}]": "value"}
```
where `key` will match `[1, 2]` and `[2, 1]`

Example script:
```
!: ALLOW RESTART
!: BOLT #VERSION#
!: AUTO RESET
!: AUTO GOODBYE

C: HELLO {"scheme": "basic", "credentials": "c", "principal": "p", }
S: SUCCESS {"server": "Neo4j/4.0.0", "connection_id": "bolt-123456789"}
{{
    C: REQ {"param1": "value1", "[optional]": "value" #SOME_PARAMETRIZATION#}
    S: RES {"param": "value", "optional_not_possible": "value" #SOME_PARAMETRIZATION#}
----
    C: REQ2 {"param_where_value_and_type_don't_matter": "*"}
    S: RES2 {"response_param_who's_value_is_*": "\*"}
    {{
        C: HEY {}
        S: HO
        S: LETS {}
        S: GO null
    ----
        C: HEY2 {"request_param_who's_value_contains_*": "some string with *"}
        S: HO2
    ----
        C: HEY3
        S: FAILURE {}
        {+
            C: RESET {}
            S: SUCCESS {}
        +}
    }}
}}
```

Limitations:
```
{{
    {{
        XXX1
    ----
        XXX2
    }}
    YYY2
----
    XXX3
}}
```
XXX must all start with "C:" (first action in any alternative branch must be
client side message). If multiple alternatives match the received message, the
first match is used. Additionally after blocks that may or may not branch at their
end (i.e. loops) a client line or the end of the script must follow.

To keep the script parser simple, `{{`, `{*`, ... need to be in a line
on their own.


robsdedude:
This will break tests in every official driver. But only because they should have been failing in the first place. The old stub server implementation just had bugs that ignored mismatches silently.

robsdedude:
What I should have mentioned is that I dropped support for `!: PORT xxxx` because I felt it doesn't belong into a script and it was never used in testkit. It wouldn't be hard to re-introduce if you feel it should be.

tests/stub/sessionparameters.py
```diff
@@ -38,7 +38,7 @@
 !: AUTO GOODBYE
 !: AUTO RESET
 
-C: RUN "RETURN 1 as n" {} {"bookmarks": ["b1", "b2"]}
+C: RUN "RETURN 1 as n" {} {"bookmarks{}": ["b1", "b2"]}
```
bigmontz, line 41: Is this change intended ?

tests/stub/routing.py
```diff
@@ -664,7 +759,7 @@ def write_tx_script_multiple_bookmarks(self):
         !: AUTO GOODBYE
         !: AUTO RESET
 
-        C: BEGIN {"db": "adb", "bookmarks": ["neo4j:bookmark:v1:tx5", "neo4j:bookmark:v1:tx29", "neo4j:bookmark:v1:tx94", "neo4j:bookmark:v1:tx56", "neo4j:bookmark:v1:tx16", "neo4j:bookmark:v1:tx68"]}
+        C: BEGIN {"db": "adb", "bookmarks{}": ["neo4j:bookmark:v1:tx5", "neo4j:bookmark:v1:tx29", "neo4j:bookmark:v1:tx94", "neo4j:bookmark:v1:tx56", "neo4j:bookmark:v1:tx16", "neo4j:bookmark:v1:tx68"]}
```
bigmontz, line 762: is it intentional ?

tests/stub/routing.py
```diff
@@ -404,7 +499,7 @@ def read_script_with_bookmarks(self):
         !: AUTO GOODBYE
         !: AUTO RESET
 
-        C: RUN "RETURN 1 as n" {} {"mode": "r", "db": "adb", "bookmarks": ["sys:1234", "foo:5678"]}
+        C: RUN "RETURN 1 as n" {} {"mode": "r", "db": "adb", "bookmarks{}": ["sys:1234", "foo:5678"]}
```
bigmontz, line 502: is it intentional ?

tests/stub/routing.py
```diff
@@ -2416,45 +2822,58 @@ def router_script_with_db_not_found_failure(self):
...
+        C: RUN "CALL dbms.routing.getRoutingTable($context, $database)" {"context": #ROUTINGCTX#, "database": "adb"} {"[mode]": "r", "db": "system", "[bookmarks]": "*"}
         C: PULL {"n": -1}
         S: SUCCESS {"fields": ["ttl", "servers"]}
         S: RECORD [1000, [{"addresses": ["#HOST#:9000"], "role":"ROUTE"}, {"addresses": ["#HOST#:9010", "#HOST#:9011"], "role":"READ"}, {"addresses": ["#HOST#:9020", "#HOST#:9021"], "role":"WRITE"}]]
         S: SUCCESS {"type": "r"}
         """
 
-    def router_script_with_bookmarks(self):
+    def router_script_with_bookmarks(self, bookmarks_in):
+        bookmarks_in = ', "bookmarks{}": %s' % json.dumps(bookmarks_in)
```
bigmontz, line 2968: is it intentional ?

tests/stub/routing.py
```diff
@@ -2808,7 +3301,7 @@ def read_script_with_bookmarks(self):
         !: AUTO HELLO
         !: AUTO GOODBYE
         !: AUTO RESET
-        C: RUN "RETURN 1 as n" {} {"mode": "r", "bookmarks": ["sys:1234", "foo:5678"]}
+        C: RUN "RETURN 1 as n" {} {"mode": "r", "bookmarks{}": ["sys:1234", "foo:5678"]}
```
bigmontz, line 3408: is it intentional ?

tests/stub/routing.py
```diff
@@ -3050,7 +3543,7 @@ def write_tx_script_multiple_bookmarks(self):
         !: AUTO HELLO
         !: AUTO GOODBYE
         !: AUTO RESET
-        C: BEGIN {"bookmarks": ["neo4j:bookmark:v1:tx5", "neo4j:bookmark:v1:tx29", "neo4j:bookmark:v1:tx94", "neo4j:bookmark:v1:tx56", "neo4j:bookmark:v1:tx16", "neo4j:bookmark:v1:tx68"]}
+        C: BEGIN {"bookmarks{}": ["neo4j:bookmark:v1:tx5", "neo4j:bookmark:v1:tx29", "neo4j:bookmark:v1:tx94", "neo4j:bookmark:v1:tx56", "neo4j:bookmark:v1:tx16", "neo4j:bookmark:v1:tx68"]}
```
bigmontz, line 3650: is it intentional ?

tests/stub/txparameters.py
```diff
@@ -42,7 +42,7 @@
 !: AUTO GOODBYE
 !: AUTO RESET
 
-C: BEGIN {"bookmarks": ["b1", "b2"]}
+C: BEGIN {"bookmarks{}": ["b1", "b2"]}
```
bigmontz, line 45: is it intentional ?

tests/stub/routing.py
robsdedude: Yes. It allows the bookmarks to be sent in arbitrary order.

tests/stub/routing.py
robsdedude: Yes. It allows the bookmarks to be sent in arbitrary order.

tests/stub/routing.py
robsdedude: Yes. It allows the bookmarks to be sent in arbitrary order.

tests/stub/routing.py
robsdedude: Yes. It allows the bookmarks to be sent in arbitrary order.

tests/stub/routing.py
robsdedude: Yes. It allows the bookmarks to be sent in arbitrary order.

tests/stub/sessionparameters.py
robsdedude: Yes. It allows the bookmarks to be sent in arbitrary order.

tests/stub/txparameters.py
robsdedude: Yes. It allows the bookmarks to be sent in arbitrary order.

bigmontz (approved):
🔧

tests/stub/iteration.py
```diff
@@ -211,11 +211,11 @@ def test_all(self):
...
 
     def test_nested(self):
-        if get_driver_name() in ['javascript']:
-            self.skipTest('It could not guarantee the order of records requests between in the nested transactions')
         # ex JAVA - java completely pulls the first query before running the second
-        if get_driver_name() not in ['go', 'dotnet', 'javascript']:
-            self.skipTest("Need support for specifying session fetch size in testkit backend")
+        if get_driver_name() in ["java"]:
+            self.skipTest(
+                "completely pulls the first query before running the second"
```
injectives, line 217: Do we know why it worked previously? Was it because the legacy stub server did not check something?

tests/stub/routing.py
```diff
@@ -56,33 +60,67 @@ def tearDown(self):
...
+    def router_script_adb(self):
         return """
         !: BOLT #VERSION#
         !: AUTO RESET
         !: AUTO GOODBYE
 
         C: HELLO {"scheme": "basic", "credentials": "c", "principal": "p", "user_agent": "007", "routing": #HELLO_ROUTINGCTX# #EXTRA_HELLO_PROPS#}
         S: SUCCESS {"server": "#SERVER_AGENT#", "connection_id": "bolt-123456789"}
-        C: ROUTE #ROUTINGCTX# [] "adb"
+        C: ROUTE #ROUTINGCTX# "*" "adb"
```
injectives, line 71: Was there some reason to make it more lenient?

tests/stub/routing.py
```diff
@@ -142,7 +198,7 @@ def router_script_with_unknown_failure(self):
 
         C: HELLO {"scheme": "basic", "credentials": "c", "principal": "p", "user_agent": "007", "routing": #HELLO_ROUTINGCTX# #EXTRA_HELLO_PROPS#}
         S: SUCCESS {"server": "#SERVER_AGENT#", "connection_id": "bolt-123456789"}
-        C: ROUTE #ROUTINGCTX# [] "adb"
+        C: ROUTE #ROUTINGCTX# [] "*"
```
injectives, line 201: Was there some need to support any database?

tests/stub/routing.py
robsdedude: Yes. This is the "default" router script that is being used in most tests. Some include bookmarks, others don't.

bigmontz: This script is used in different test scenarios which are not testing bookmarks, maybe It's because of this.

tests/stub/routing.py
```diff
@@ -219,7 +270,7 @@ def router_script_with_one_reader_and_exit(self):
...
```
injectives, line 273: Was there some need to support any database?

tests/stub/iteration.py
robsdedude: I think it didn't work before. `if get_driver_name() not in ['go', 'dotnet', 'javascript']:` disabled the test for Python and Java, admittedly in a somewhat convoluted way.

tests/stub/routing.py
robsdedude: Indeed. Additionally, there are dedicated tests for checking if the driver sends bookmarks when fetching the RT.

tests/stub/routing.py
robsdedude: Yup, `test_should_fail_discovery_when_router_fails_with_unknown_code` calls `verifyConnectivity` on that router. So in that case the DB is either `system` or `null`, depending on the driver's implementation of `verifyConnectivity`. Meanwhile, `test_should_request_rt_from_all_initial_routers_until_successful` uses that router script with a session on `adb`.

tests/stub/routing.py
robsdedude: `test_should_use_resolver_during_rediscovery_when_existing_routers_fail` uses this router for a session on `adb` while `test_should_successfully_acquire_rt_when_router_ip_changes` uses it to `verifyConnectivity`. Side note: I'm not a fan of (ab)using `verifyConnectivity` to force the driver to fetch a new RT because a) I'm not sure we ever specified that function to do that, b) at least the JS driver does something different, and c) the method's name has nothing to do with routing tables, why should one assume it always and only updates the RT?. I think in the long run, we should have another method exposed by the testkit backend to do this.

tests/stub/routing.py
```diff
@@ -860,74 +963,33 @@ def work(tx):
         self._readServer1.done()
         self.assertEqual([[1]], sequences)
 
-    def test_should_round_robin_readers_when_reading_using_session_run(self):
```
injectives, line 863: These 2 tests were ported from java driver. I know they are very implementation specific, but we could just make them a bit more lenient and basically check if one of the servers gets used and goes down then the driver uses the other server. However, we may already have such tests.

tests/stub/routing.py
```diff
@@ -2011,12 +2276,12 @@ def test_should_ignore_system_bookmark_when_getting_rt_for_multi_db(self):
 
     def test_should_request_rt_from_all_initial_routers_until_successful(self):
         # TODO add support and remove this block
-        if get_driver_name() in ['dotnet', 'javascript']:
+        if get_driver_name() in ['dotnet']:
             self.skipTest("needs ROUTE bookmark list support")
-        if get_driver_name() in ['python', 'javascript', 'go', 'dotnet']:
+        if get_driver_name() in ['go', 'dotnet']:
```
injectives, line 2385: Good to see that python and javascript implement DNS discovery properly 👍

tests/stub/routing.py
```diff
@@ -2031,11 +2296,11 @@ def test_should_request_rt_from_all_initial_routers_until_successful(self):
...
         # The resolver is used to convert the original address to multiple fake domain names.
         def resolver(address):
-            nonlocal resolver_call_num
+            nonlocal resolver_calls
             nonlocal resolved_addresses
-            self.assertEqual(0, resolver_call_num)
-            self.assertEqual(self._routingServer1.address, address)
-            resolver_call_num += 1
+            resolver_calls[address] = resolver_calls.get(address, 0) + 1
+            if address != self._routingServer1.address:
```
injectives, line 2406: What do we need this `if` statement for?

tests/stub/routing.py
robsdedude: :tada:

tests/stub/routing.py
robsdedude: We have some tests that go in that direction. The biggest problem is, how do make sure that the first server the driver contacts will be the broken one? This would require some sort of callback or communication between the servers, I think.

tests/stub/routing.py
injectives: While I would not use it for testing bookmarks, I do like when scripts match my expectation as close as possible.

tests/stub/routing.py
injectives: This is simply the legacy of java driver. When those tests were introduced there, they probably did not expect it to be moved to testkit one day... It can and should be improved later on :)

tests/stub/routing.py
robsdedude: The Python driver calls the custom resolver function for the initial address as well as for each address that is received as part of a routing table ([Trello card](https://trello.com/c/HhizyhRm/3495)). This behavior will not be unified before 5.0 because it's a breaking change.

tests/stub/routing.py
injectives: Sorry, I did not mean breaking the server. I just expected the server to finish the script and exit, which would leave the second server running. But if you believe there are tests that already cover this case, then it would just be a duplicate :)

injectives (approved):
LGTM
I did not have time to review every line, but I checked the most relevant parts.

tests/stub/routing.py
robsdedude: But why shouldn't a driver be allowed to contact the same server twice?

PR #74, opened 2021-03-30
Adding support for bookmarks in ROUTE message

gjmwoods:


bigmontz:
Is it already implemented on the server? 

What's the effect of the bookmark in the route message (server side)?

gjmwoods:
yes almost, just adding tests for parsing the bookmark,
The solution is to solve a case where you create a database (say `foo`) and then wish to execute a query against that db. You run the create in a separate session and then use the bookmark from that session in the subsequent session to execute the query against `foo`. This should have the effect of triggering a new ROUTE call to the server to get the RT for that new DB and the bookmark is used to wait for the creation on the DB to finish (as CREATE database) is asynchronous.

See: https://github.com/neo4j-drivers/testkit/blob/97f97f6e8f42c5d5eb693ca227da37f87832412c/tests/stub/routing.py#L781

bigmontz:
That's a really specific case. Nice catch.

tests/stub/routing.py
```diff
@@ -1398,7 +1521,7 @@ def work(tx):
 
     def test_should_use_initial_router_for_discovery_when_others_unavailable(self):
         # TODO add support and remove this block
-        if get_driver_name() in ['python', 'javascript', 'go', 'dotnet']:
+        if get_driver_name() in ['python', 'javascript', 'go', 'dotnet', 'java']:
```
injectives: Please remember to remove this in the next PR.

injectives (approved):
LGTM

tests/stub/routing.py
gjmwoods: I also added this one explicitly

PR #76, opened 2021-04-01
Removing the javascript conditionals

bigmontz:
The conditions were moved to the driver repository

### DIFF
--- a/tests/stub/transport.py
+++ b/tests/stub/transport.py
@@ -38,7 +38,6 @@ class Transport(TestkitTestCase):
         self._session = self._driver.session("w")
 
     def tearDown(self):
-        self._backend.close()
         self._server.reset()
         super().tearDown()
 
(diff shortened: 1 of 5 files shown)


tests/stub/transport.py
```diff
@@ -38,13 +38,9 @@ def setUp(self):
         self._session = self._driver.session("w")
 
     def tearDown(self):
-<<<<<<< HEAD
         self._backend.close()
```
robsdedude: Oops, how did that merge conflict end up here :fearful: 
`self._backend.close()` can also go, as it's part of `TestkitTestCase.tearDown`.

robsdedude (changes requested):

tests/stub/transport.py
bigmontz: Sure, `self._backend.close()` removed.

robsdedude (approved):
LGTM

PR #79, opened 2021-04-08
Routing: Fix try_count verification

bigmontz:
The drivers differs a bit about when it should or not call the transaction function.
The java driver only call the transaction function if the connection is fully setup, meanwhile the JS driver delay the connection resolution until the last moment.

Thus the try_count is not a good way to indirect check if the driver try to call all the servers needed. Ideally, It should have a way to check if the full script was called, but since it's not possible at this point, the adopt alternative was checking for the occurrence of some events which could indicates the script was run on the stub server execution log.

### DIFF
--- a/tests/stub/routing.py
+++ b/tests/stub/routing.py
@@ -1128,8 +1128,17 @@ class Routing(TestkitTestCase):
         self._routingServer1.done()
         self._writeServer1.done()
         self._writeServer2.done()
+
+        self.assertGreaterEqual(
+            self._routingServer1.count_responses("SUCCESS"), 2
+        )
+        self.assertGreaterEqual(
+            self._writeServer1.count_responses("<EXIT>"), 1
+        )
+        self.assertGreaterEqual(
+            self._writeServer2.count_responses("SUCCESS"), 4
+        )
         self.assertEqual([[], []], sequences)
-        self.assertEqual(2, num_retries)
 
     def test_should_fail_when_writing_on_unexpectedly_interrupting_writer_using_session_run(self):
         # TODO remove this block once all languages work
@@ -1564,8 +1573,23 @@ class Routing(TestkitTestCase):
         self._readServer1.done()
         self._readServer2.done()
         self._readServer3.done()
+
+        self.assertGreaterEqual(
+            self._routingServer1.count_responses("SUCCESS"), 2
+        )
+        self.assertGreaterEqual(
+            self._routingServer2.count_responses("SUCCESS"), 2
+        )
+        self.assertGreaterEqual(
+            self._readServer1.count_responses("<EXIT>"), 1
+        )
+        self.assertGreaterEqual(
+            self._readServer2.count_responses("RECORD"), 1
+        )
+        self.assertGreaterEqual(
+            self._readServer3.count_responses("<EXIT>"), 1
+        )
         self.assertEqual([[1]], sequences)
-        self.assertEqual(3, try_count)
 
     def test_should_retry_write_tx_and_rediscovery_until_success(self):
         # TODO remove this block once all languages work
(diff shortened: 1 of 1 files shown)


tests/stub/routing.py
```diff
@@ -1128,8 +1128,14 @@ def work(tx):
...
         self._writeServer2.done()
+
+        self.assertGreaterEqual(
+            self._routingServer1.count_responses("SUCCESS"), 2)
+        self.assertGreaterEqual(
+            self._writeServer1.count_responses("<EXIT>"), 1)
+        self.assertGreaterEqual(
+            self._writeServer2.count_responses("SUCCESS"), 4)
         self.assertEqual([[], []], sequences)
-        self.assertEqual(2, num_retries)
```
injectives, line 1132: Should we also remove the `num_retries` declaration?

robsdedude:
You can check if the whole stub script was played. `server.done()` will raise an exception if there is the script is unfinished.
Note: if `<EXIT>` is invoked at any point in the script, this counts as finishing the script. Especially noteworthy, it `!AUTO: GOODBYE` is present, the client can reach the end of the script at any time by sending a `GOODBYE` message.
Furthermore, for `!ALLOW RESTART` and `!ALLOW CONCURRENT`, `server.done()` will check that all open connections have reached the end of the script. This means, if the client never connects to the server, this also counts as having finished the script.

tests/stub/routing.py
```diff
@@ -1128,8 +1128,14 @@ def work(tx):
         self._routingServer1.done()
         self._writeServer1.done()
         self._writeServer2.done()
+
+        self.assertGreaterEqual(
+            self._routingServer1.count_responses("SUCCESS"), 2)
+        self.assertGreaterEqual(
+            self._writeServer1.count_responses("<EXIT>"), 1)
+        self.assertGreaterEqual(
+            self._writeServer2.count_responses("SUCCESS"), 4)
```
robsdedude:
```
        self.assertGreaterEqual(
            self._routingServer1.count_responses("SUCCESS"), 2
        )
```
...

tests/stub/routing.py
```diff
@@ -1564,8 +1570,13 @@ def work(tx):
         self._readServer1.done()
         self._readServer2.done()
         self._readServer3.done()
+
+        self.assertGreaterEqual(self._routingServer1.count_responses("SUCCESS"), 2)
+        self.assertGreaterEqual(self._routingServer2.count_responses("SUCCESS"), 2)
+        self.assertGreaterEqual(self._readServer1.count_responses("<EXIT>"), 1)
+        self.assertGreaterEqual(self._readServer2.count_responses("RECORD"), 1)
+        self.assertGreaterEqual(self._readServer3.count_responses("<EXIT>"), 1)
```
robsdedude: Try to keep the lines shorter or equal to 80 characters long.

tests/stub/routing.py
```diff
@@ -1599,8 +1610,12 @@ def work(tx):
         self._writeServer1.done()
         self._writeServer2.done()
         self._writeServer3.done()
+        self.assertGreaterEqual(self._routingServer1.count_responses("SUCCESS"), 2)
+        self.assertGreaterEqual(self._routingServer2.count_responses("SUCCESS"), 2)
+        self.assertGreaterEqual(self._writeServer1.count_responses("<EXIT>"), 1)
+        self.assertGreaterEqual(self._writeServer2.count_responses("SUCCESS"), 4)
+        self.assertGreaterEqual(self._writeServer3.count_responses("<EXIT>"), 1)
```
robsdedude: 80 characters

robsdedude (changes requested):

bigmontz:
> You can check if the whole stub script was played. `server.done()` will raise an exception if there is the script is unfinished.
> Note: if `<EXIT>` is invoked at any point in the script, this counts as finishing the script. Especially noteworthy, it `!AUTO: GOODBYE` is present, the client can reach the end of the script at any time by sending a `GOODBYE` message.
> Furthermore, for `!ALLOW RESTART` and `!ALLOW CONCURRENT`, `server.done()` will check that all open connections have reached the end of the script. This means, if the client never connects to the server, this also counts as having finished the script.

Yes, we have to make sure also if the client connects to the server. Maybe I could simplify by just check if the server replied server SUCCESS once, for example.

robsdedude:
> Yes, we have to make sure also if the client connects to the server. Maybe I could simplify by just check if the server replied server SUCCESS once, for example.

For asserting that the driver connected exactly once to a server that has `!ALLOW RESTART` or `!ALLOW CONCURRENT` in its script, I recommend something like this

```
self._routingServer1.done()
self.assertLessEqual(self._routingServer1.count_responses("<ACCEPT>"), 1)
```

Alternatively, one can consider whether the `!ALLOW` bang line(s) can be removed. Then, the second line is superfluous.

bigmontz:
The premises of the changes were change, another PR will be open resolving the issue in another way.

PR #87, opened 2021-04-13
Enabled 2 tests for java driver where router is also a reader

injectives:
The java driver uses separate connections for routers and readers/writers when readers/writers are addressed by domain names in routing table. However, it will aim to use the same connection if IP addresses are used and there is no ambiguity involved. Since the purpose of these tests in not to test DNS resolution, it is suggested to switch them to IP-based addressing.

### DIFF
--- a/tests/stub/routing.py
+++ b/tests/stub/routing.py
@@ -36,7 +36,8 @@ class Routing(TestkitTestCase):
         self._writeServer2 = StubServer(9021)
         self._writeServer3 = StubServer(9022)
         self._uri_template = "neo4j://%s:%d"
-        self._uri_template_with_context = self._uri_template + "?region=china&policy=my_policy"
+        self._uri_template_with_context = \
+            self._uri_template + "?region=china&policy=my_policy"
         self._uri_with_context = self._uri_template_with_context % (
             self._routingServer1.host, self._routingServer1.port)
         self._auth = types.AuthorizationToken(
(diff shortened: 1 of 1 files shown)


tests/stub/routing.py
```diff
@@ -1598,13 +1608,19 @@ def test_should_send_empty_hello(self):
...
+        if platform == "linux":
+            ip_addresses = self.get_ip_addresses()
+        if len(ip_addresses) < 1:
+            self.skipTest("only linux is supported at the moment")
+        ip_address = ip_addresses[0]
+        driver = Driver(self._backend, self._uri_template % (ip_address, self._routingServer1.port), self._auth,
+                        self._userAgent)
         self._routingServer1.start(script=self.router_script_with_empty_context_and_reader_support(),
-                                   vars=self.get_vars())
+                                   vars=self.get_vars(ip_address))
```
robsdedude: Keyword argument.

tests/stub/routing.py
```diff
@@ -1571,12 +1575,18 @@ def test_should_successfully_read_from_readable_router_using_tx_function(self):
...
+        # IP-based address model.
+        ip_addresses = []
+        if platform == "linux":
+            ip_addresses = self.get_ip_addresses()
+        if len(ip_addresses) < 1:
+            self.skipTest("only linux is supported at the moment")
+        ip_address = ip_addresses[0]
+        driver = Driver(self._backend, self._uri_template_with_context % (ip_address, self._routingServer1.port),
+                        self._auth, self._userAgent)
+        self._routingServer1.start(script=self.router_script_with_reader_support(), vars=self.get_vars(ip_address))
```
robsdedude: Prefer to use `self.get_vars(host=ip_address)`. Explicit keyword arguments make refactoring easier.

tests/stub/routing.py
```diff
@@ -1571,12 +1575,18 @@ def test_should_successfully_read_from_readable_router_using_tx_function(self):
...
         if get_driver_name() in ['python']:
             self.skipTest("requires investigation")
-        # TODO remove this block once it works
-        if get_driver_name() in ['java']:
-            self.skipTest("java driver uses separate connections for routing and reading, "
-                          "but the stub server does not currently support this")
-        driver = Driver(self._backend, self._uri, self._auth, self._userAgent)
-        self._routingServer1.start(script=self.router_script_with_reader_support(), vars=self.get_vars())
+        # Some drivers (for instance, java) may use separate connections for readers and writers when they are addressed
+        # by domain names in routing table. Since this test is not for testing DNS resolution, it has been switched to
```
robsdedude: Python style guide recommends 80 characters max. per line.

robsdedude (approved):
It is not the cleanest solution but the best with the current feature set of the stubserver. Except for code style, this lgtm.

tests/stub/routing.py
injectives: Updated :)

tests/stub/routing.py
injectives: Updated :)

tests/stub/routing.py
injectives: I've reformatted the 2 methods and the urls declaration. I decided to keep the other parts unchanged. If I reformat the whole file, this PR will become unreadable. I am happy to do this separately, but I assume your upcoming changes will bring proper formatting anyway.

injectives:
> It is not the cleanest solution but the best with the current feature set of the stubserver. Except for code style, this lgtm.

Thanks 👍 I used what we have at the moment as having tests running is preferable, but I hope this can be improved later once we have more features in the stub server.

PR #91, opened 2021-04-26
Make tests more lenient to let Python driver pass

robsdedude:
Most of theses changes originate in the Python driver calling the custom resolver for addresses received in routing tables as well (cf. [Trello card](https://trello.com/c/HhizyhRm/3495)).

### DIFF
--- a/tests/stub/routing.py
+++ b/tests/stub/routing.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 try:
     import fcntl
 except ImportError:  # e.g. on Windows
@@ -2008,6 +2009,12 @@ class Routing(TestkitTestCase):
                                  database=self.get_db())
         sequences = []
         self._routingServer1.done()
+        try:
+            driver.verifyConnectivity()
+        except types.DriverError:
+            # make sure the driver noticed that its old connection to
+            # _routingServer1 is dead
+            pass
         self._routingServer1.start(script=self.router_script_adb(),
                                    vars=self.get_vars())
 
@@ -2187,9 +2194,14 @@ class Routing(TestkitTestCase):
         if get_driver_name() in ['go', 'dotnet']:
             self.skipTest("resolver not implemented in backend")
 
+        resolver_calls = defaultdict(lambda: 0)
+
         def resolver(address):
-            self.assertEqual(self._routingServer1.address, address)
-            return [self._routingServer1.address, self._routingServer3.address]
+            resolver_calls[address] += 1
+            if address == self._routingServer1.address:
+                return [self._routingServer1.address,
+                        self._routingServer3.address]
+            return [address]
 
         driver = Driver(self._backend, self._uri_with_context, self._auth,
                         self._userAgent, resolver)
@@ -2222,6 +2234,30 @@ class Routing(TestkitTestCase):
         self._routingServer3.done()
         self._readServer1.done()
         self.assertEqual([[1]], sequences)
+        if len(resolver_calls) == 1:
+            # driver that calls resolver function only on initial router address
+            self.assertEqual(resolver_calls.keys(),
+                             {self._routingServer1.address})
+            # depending on whether the resolve result is treated equally to a
+            # RT table entry or is discarded after an RT has been retrieved
+            # successfully.
+            self.assertIn(resolver_calls[self._routingServer1.address], {1, 2})
+        else:
+            fake_reader_address = self._routingServer1.host + ":9099"
+            # driver that calls resolver function for ever address (initial
+            # router and server addresses returned in routing table
+            self.assertLessEqual(resolver_calls.keys(),
+                                 {self._routingServer1.address,
+                                  fake_reader_address,
+                                  self._routingServer2.address,
+                                  self._readServer1.address,
+                                  # readServer2 isn't part of this test but is
+                                  # in the RT of router_script_adb by default
+                                  self._readServer2.address})
+            self.assertIn(resolver_calls[self._routingServer1.address], {1, 2})
+
+            self.assertEqual(resolver_calls[fake_reader_address], 1)
+            self.assertEqual(resolver_calls[self._readServer1.address], 1)
 
     def should_support_multi_db(self):
         return True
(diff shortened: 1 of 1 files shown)


tests/stub/routing.py
```diff
@@ -2008,6 +2009,12 @@ def test_should_accept_routing_table_without_writers_and_then_rediscover(self):
                                  database=self.get_db())
         sequences = []
         self._routingServer1.done()
+        try:
+            driver.verifyConnectivity()
+        except types.DriverError:
+            # make sure the driver noticed that its old connection to
+            # _routingServer1 is dead
```
injectives: I can see the comment, but I am not sure why the driver needs to notice this... Could you explain please? (I know this test currently relies on the driver fetching RT in the `verify` method, which is suboptimal.)

tests/stub/routing.py
```diff
@@ -2222,6 +2234,30 @@ def work(tx):
...
         self._readServer1.done()
         self.assertEqual([[1]], sequences)
+        if len(resolver_calls) == 1:
+            # driver that calls resolver function only on initial router address
+            self.assertEqual(resolver_calls.keys(),
+                             {self._routingServer1.address})
+            # depending on whether the resolve result is treated equally to a
+            # RT table entry or is discarded after an RT has been retrieved
+            # successfully.
+            self.assertIn(resolver_calls[self._routingServer1.address], {1, 2})
```
injectives: Out if interest, which drivers would get `1` and `2` here?

tests/stub/routing.py
```diff
@@ -2222,6 +2234,30 @@ def work(tx):
...
+        else:
+            fake_reader_address = self._routingServer1.host + ":9099"
+            # driver that calls resolver function for ever address (initial
```
injectives: Did you mean `every`? :)

tests/stub/routing.py
```diff
@@ -2222,6 +2234,30 @@ def work(tx):
...
+            # router and server addresses returned in routing table
+            self.assertLessEqual(resolver_calls.keys(),
+                                 {self._routingServer1.address,
+                                  fake_reader_address,
+                                  self._routingServer2.address,
+                                  self._readServer1.address,
+                                  # readServer2 isn't part of this test but is
+                                  # in the RT of router_script_adb by default
+                                  self._readServer2.address})
+            self.assertIn(resolver_calls[self._routingServer1.address], {1, 2})
```
injectives: Out if interest, which drivers would get `1` and `2` here?

tests/stub/routing.py
```diff
@@ -2255,7 +2291,10 @@ def test_should_read_successfully_on_empty_discovery_result_using_session_run(se
             self.skipTest("resolver not implemented in backend")
 
         def resolver(address):
-            return [self._routingServer1.address, self._routingServer2.address]
+            if address == self._routingServer1.address:
+                return (self._routingServer1.address,
+                        self._routingServer2.address)
+            return address,
```
injectives, line 2291: Is `,` needed here?

tests/stub/routing.py
```diff
@@ -2756,11 +2795,15 @@ def router_script_with_one_reader_and_exit(self):
...
-
         !: AUTO GOODBYE
+
         C: HELLO {"scheme": "basic", "credentials": "c", "principal": "p", "user_agent": "007", "routing": #HELLO_ROUTINGCTX# #EXTRA_HELLO_PROPS# }
         S: SUCCESS {"server": "#SERVER_AGENT#", "connection_id": "bolt-123456789"}
-        C: RUN "CALL dbms.routing.getRoutingTable($context, $database)" {"context": #ROUTINGCTX#, "database": "*"} {"[mode]": "r", "db": "system"}
+        {{
+            C: RUN "CALL dbms.routing.getRoutingTable($context)" {"context": #ROUTINGCTX#} {"[mode]": "r", "db": "system"}
+        ----
+            C: RUN "CALL dbms.routing.getRoutingTable($context, $database)" {"context": #ROUTINGCTX#, "database": "*"} {"[mode]": "r", "db": "system"}
```
injectives, line 2799: Is there some difference in how drivers invoke this? I thought that the previous version worked...

tests/stub/routing.py
robsdedude: Yes, to make it a tuple. The resolver function is supposed to return a tuple or list.

tests/stub/routing.py
robsdedude: I'm pretty sure, some drivers use the default db (can explicitly pass `null` or omit the database argument), while other's use the system db to do `verifyConnectivity`. Furthermore, this is just a missed change from the big PR #70. `router_script_with_one_reader_and_exit` in `Routing` has this line `C: ROUTE #ROUTINGCTX# [] "*"`, so this is the equivalent for Bolt 4.1. And finally, this script is used in `test_should_use_resolver_during_rediscovery_when_existing_routers_fail` and `test_should_successfully_acquire_rt_when_router_ip_changes` both seem not to be about testing the behavior of `verifyConnectivity`.

tests/stub/routing.py
robsdedude: oops :sweat_smile:

tests/stub/routing.py
robsdedude: Tried it again, works. I think it was a left-over from when the Python driver implemented the `verifyConnectivity` method in a slightly different way :shrug:

tests/stub/routing.py
robsdedude: .net, js, and go, skip this test, only java reaches this branch and calls the resolver twice for `_routingServer1`. However, if we decide to switch the the routing/resolving algorithm I proposed, this will be reduced to a single call.

tests/stub/routing.py
robsdedude: .net, js, and go, skip this test, only python reaches this branch and calls the resolver twice for _routingServer1.

tests/stub/routing.py
robsdedude: Actually, no. It wouldn't. It should always be two. I'll change that.

tests/stub/routing.py
injectives: As discussed in another chat, this is due to the difference in the way drivers implement the `verifyC` method. Please make sure this difference is captured somewhere in the cards so we can address it later on.

injectives (approved):
LGTM

tests/stub/routing.py
robsdedude: @injectives turns out it was needed indeed. Must have missed it when testing again. The reason is that the Python driver keeps the connection to the router alive in the pool. Then `session.run("RETURN 1 as n").consume()` will cause the driver to try to re-fetch the RT because it's not knowing any writers. Since this in an auto-commit tx, there is no retry. The pool connection will fail, the tx, will fail, the test will fail.

PR #94, opened 2021-04-28
Added authorization expired response tests

injectives:
These tests verify that the driver handles the authorization expired response correctly. Such response can be returned by the server when authorization info that is maintained on the server side expires.

### DIFF
--- a/tests/stub/authorization.py
+++ b/tests/stub/authorization.py
@@ -1,0 +1,634 @@
+from nutkit.frontend import Driver
+import nutkit.protocol as types
+from tests.shared import (
+    get_driver_name,
+    TestkitTestCase,
+)
+from tests.stub.shared import StubServer
+
+
+def get_extra_hello_props():
+    if get_driver_name() in ["java"]:
+        return ', "realm": ""'
+    elif get_driver_name() in ["javascript"]:
+        return ', "realm": "", "ticket": ""'
+    return ""
+
+
+class AuthorizationTests(TestkitTestCase):
+    def setUp(self):
+        super().setUp()
+        self._routingServer1 = StubServer(9000)
+        self._readServer1 = StubServer(9010)
+        self._readServer2 = StubServer(9011)
+        self._uri = "neo4j://%s:%d" % (self._routingServer1.host,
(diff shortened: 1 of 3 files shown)


injectives:
@gjmwoods please check that the scripts match the server-side behaviour.

injectives:
The Java driver is expected to fail until we have merged this: https://github.com/neo4j/neo4j-java-driver/pull/886

gjmwoods (approved):
looks like a very comprehensive suite of tests!

tests/stub/authorization.py
```diff
@@ -0,0 +1,634 @@
...
+
+def get_extra_hello_props():
+    if get_driver_name() in ["java"]:
+        return ', "realm": ""'
+    elif get_driver_name() in ["javascript"]:
+        return ', "realm": "", "ticket": ""'
+    return ""
+
+
+class Tests(TestkitTestCase):
```
robsdedude: Maybe give it a more descriptive name like `AuthorizationTests` or such.

injectives: Good point, I have updated the PR, thank you 👍

tests/stub/routing.py
```diff
@@ -129,8 +129,8 @@ def router_with_bookmarks_script_create_adb(self):
         C: HELLO {"scheme": "basic", "credentials": "c", "principal": "p", "user_agent": "007", "routing": #HELLO_ROUTINGCTX# #EXTRA_HELLO_PROPS#}
         S: SUCCESS {"server": "#SERVER_AGENT#", "connection_id": "bolt-123456789"}
         C: BEGIN {"db": "system"}
-        C: RUN "CREATE database foo" {} {}
         S: SUCCESS {}
+        C: RUN "CREATE database foo" {} {}
```
robsdedude, line 133: I think I've seen many of those, where the stub server expects the driver to pipeline multiple requests. We should probably (in another PR) change that.

robsdedude (approved):
I haven't looked in great detail at every singe test, but overall they look good to me. If there are minor problems, I'll probably discover them while adjusting the Python driver.

PR #95, opened 2021-04-29
Restructure routing tests

robsdedude:
This is a proof of concept of how we could re-structure the tests to be easier to maintain. I constantly got lost in the nearly 4k LOC long routing test file.


boltstub/parsing.py
```diff
@@ -970,7 +970,8 @@ def start(self, tree):
     def block_list(self, tree):
         blocks = []
         for child in tree.children:
-            if isinstance(child, lark.Token):
+            if ((isinstance(child, lark.Tree) and child.data == "comment_block")
+                    or isinstance(child, lark.Token)):
```
robsdedude: Leftover from when comments were block. Can probably be removed.

tests/stub/routing/_routing.py
```diff
@@ -0,0 +1,154 @@
...
+        return ', "realm": ""'
+    elif get_driver_name() in ["javascript"]:
+        return ', "realm": "", "ticket": ""'
+    return ""
+
+
+# This should be the latest/current version of the protocol.
+# Older protocol that needs to be tested inherits from this and override
+# to handle variations.
+class RoutingBase(TestkitTestCase):
```
bigmontz, line 34: Maybe it could be called `RoutingCurrent` or `RoutingLatest` since the description is `should be the latest/current version of the protocol`

robsdedude: I was thinking about it :thinking:

bigmontz (approved):
I think It's going in good direction.

The amount of different things being tested in the RoutingV4x3 and so on stills really big also after the changes. I don't think it should be addressed in this PR, but I think we could somehow group these tests by context in the future. 

🥇

robsdedude:
> The amount of different things being tested in the RoutingV4x3 and so on stills really big also after the changes. I don't think it should be addressed in this PR, but I think we could somehow group these tests by context in the future.

I totally agree.

PR #99, opened 2021-04-29
Added optional RESETs to successful auth scripts

injectives:
This is useful for drivers that send a RESET when releasing connection back to connection pool and prevents unwanted exceptions in the logs.


bigmontz (approved):
🦖

PR #103, opened 2021-04-30
Adjust and clean-up authorization tests for Python

robsdedude:


injectives:
This WIP looks promising 👍
I felt like we needed some form of callback in between TX function retries, but I would surface this API for TX functions until we see valid use-cases in other areas. In my view, it would also be nice to register hooks through some methods instead of adding things to maps, but this is just my subjective view.

injectives:
Given that it is possible to detect when first attempt happens, would it be possible to use 1 stub server for TX function tests? (I haven't tried this)

robsdedude:
> Given that it is possible to detect when first attempt happens, would it be possible to use 1 stub server for TX function tests? (I haven't tried this)

Yes, it probably is, and this should be preferable because it doesn't require the driver to support routing :+1:

robsdedude:
> > Given that it is possible to detect when first attempt happens, would it be possible to use 1 stub server for TX function tests? (I haven't tried this)
> 
> Yes, it probably is, and this should be preferable because it doesn't require the driver to support routing

But at the same time I'd like to leave this work for another PR as I want this to be merged before doing the Python driver beta release

injectives (approved):
LGTM

PR #104, opened 2021-04-30
Check connections for closure on LDAP Auth Credential expiration

bigmontz:


tests/stub/authorization.py
```diff
@@ -655,3 +655,70 @@ def get_vars(self, host=None):
...
+        super().tearDown()
+
+    def read_return_1_failure_return_2_succeed_script(self):
+        return """
+        !: BOLT 4
+        !: AUTO HELLO
+        !: AUTO GOODBYE
+        !: AUTO RESET
+        !: ALLOW RESTART
+        !: ALLOW CONCURRENT
```
robsdedude: `!: ALLOW CONCURRENT` implies `!: ALLOW RESTART`. No need but also no harm in having it there explicitly.

tests/stub/authorization.py
```diff
@@ -655,3 +657,69 @@ def get_vars(self, host=None):
...
+            session2.run('RETURN 1 as n').next()
+        except types.DriverError as e:
+            self.assert_is_authorization_error(e)
+
+        session2.close()
+        session1.close()
+        accept_count = self._readServer.count_responses("<ACCEPT>")
+        hangup_count = self._readServer.count_responses("<HANGUP>")
+        
+        self.assertEqual(accept_count, hangup_count)
```
injectives, line 804: This needs some refinement as it very likely to fail on the java driver, which does not hangup on all connections immediately, but makes sure that they do net get reused. In other words, the java driver performs a lazy cleanup on connection acquisition.

bigmontz: So, maybe it could be the case load the accept count at this stage. Then create a new session (run a query to force the connection be fetched, JS stuff :D ) and then count fetch the HANGUPs to compare. 

I will try to do this refinement and run against the JS driver (with and without the fix)

bigmontz: @injectives I updated the test with a more lenient implementation.

nutkit/protocol/responses.py
```diff
@@ -23,7 +23,9 @@
 
 class RunTest:
     """ Response to StartTest indicating that the test can be started"""
-    pass
+
+    def __init__(self, capabilities=[]):
```
robsdedude: This is a bad idea because lists are mutable and that default argument will be the same list for every call.

```python
def __init__(self, capabilities=None):
    if capabilities is None:
        capabilities = []
    ...
```

tests/shared.py
```diff
@@ -29,6 +29,22 @@ def new_backend():
     return Backend(host, port)
 
 
+def driver_feature(capability):
+    def get_valid_test_case(*args, **kwargs):
+        if not isinstance(args[0], TestkitTestCase):
```
robsdedude:
```python
if not args or not isinstance(args[0], TestkitTestCase):
```
To handle the case if someone decorates a function that is called with no arguments.

robsdedude (changes requested):
Generally, I like the idea. We should consider two more things:
 - We need a place to document all possible capabilities. Better yet, have them all as constants  in one place in the code base
  - I'd prefer if testkit would ask the backend actively for the required capabilities. This way we greatly reduce the needed network throughput as the backend does not have to send the full list of capabilities for each test.

nutkit/protocol/responses.py
bigmontz: Thanks, addressed.

tests/shared.py
bigmontz: Thanks, addressed.

AndyHeap-NeoTech:
This all looks ok to me.

tests/feature.py
```diff
@@ -0,0 +1,8 @@
+"""
+Enumerate all the capabilities in the drivers
+"""
+from enum import Enum
+
+
+class Feature(Enum):
+    AUTHORIZATION_EXPIRED_TREATMENT = 'AutorizationExpiredTreament'
```
robsdedude: I'd argue that that should go somewhere in `nutkit/protocol/responses.py` or similar, as those stings are sent over the wire, aren't they?

tests/shared.py
```diff
@@ -40,6 +87,7 @@ def setUp(self):
         id_ = re.sub(r"^([^\.]+\.)*?tests\.", "", self.id())
         self._backend = new_backend()
         self.addCleanup(self._backend.close)
+        self._driver_features = get_driver_features(self._backend)
```
robsdedude, line 100: This approach still makes the back end send the full list of it's features if I read the code correctly. Even more so, it requests the list even for tests that don't even need to know them.

robsdedude (changes requested):

tests/shared.py
bigmontz: It didn't, the `get_driver_features` func only goes over the wire once and then cache the result.

tests/feature.py
bigmontz: The strings are sent over the wire, I could move it to the protocol. But this one is kind of grey area between the protocol and the test context.

tests/shared.py
robsdedude: I just noticed the cache. Guess that's ok, but it assumes that the backend is always the same during execution... Seems ok to me. But if you have the features in a global place like that, why do you even bother saving and reading them from `TestkitTestCase._driver_features`?

tests/feature.py
robsdedude: I've seen many tests that do `import nutkit.protocol as types`. So it wouldn't be the first place where tests know about the protocol's classes.

tests/shared.py
```diff
@@ -30,6 +32,51 @@ def new_backend():
...
+features = None
+
+
+def get_driver_features(backend):
+    global features
+    if features is None:
+        # TODO Remove when dotnet implements the GetFeature message
+        if get_driver_name() in ['dotnet']:
+            warnings.warn("""Driver does not implements GetFeatures.
+            """)
```
robsdedude: Is there a specific reason for this to be a multi-line sting? Especially given the other warning is a regular string?

bigmontz: It was a leftover. I will fix it.

tests/shared.py
bigmontz: (Just for the record)

It's because I don't anyone depends on the feature variable, since it's meant for internal usage inside the `get_driver_features`, to isolate the code a bit.

tests/stub/authorization.py
injectives: The changes look better. 👍  I have actually implemented the feature list support in Java Driver: https://github.com/neo4j/neo4j-java-driver/pull/889
However, I am slightly concerned about the accept and hangup counts because I get either `1` or `2` for both. Obviously, the test passes in both cases but it seems like we should find an answer why it happens first.
On a separate note, does the new stub server always forcibly hangup when it returns the auth expired error? (this is would match the real server behaviour)

nutkit/protocol/feature.py
```diff
@@ -0,0 +1,8 @@
+"""
+Enumerate all the capabilities in the drivers
+"""
+from enum import Enum
+
+
+class Feature(Enum):
+    AUTHORIZATION_EXPIRED_TREATMENT = 'AutorizationExpiredTreament'
```
injectives:
```suggestion
    AUTHORIZATION_EXPIRED_TREATMENT = 'AuthorizationExpiredTreatment'
```

tests/stub/authorization.py
bigmontz: It's supposed to always be at least two connections to hangup, since it is running two sessions in parallel. But i don't know if it have more than 2 connections open and then closed will be an issue for this test in particular. 

I will add the exit after the exit on the server failure, to make it closer to the real world behaviour.

injectives: While it looks like it should run 2 sessions in parallel, this is not necessarily the case. For instance, the Java driver is slightly eager and tries looking ahead, which causes it to realise that the whole data has been consumed and there is nothing to do in `session1` and it releases the connection. This sometimes results in `session1` and `session2` effectively using the same connection and introduces an unexpected/unanticipated variation to the test.

This may apply to any driver that tries looking ahead. I managed to get it to work as expected with the following changes:
1. Update the script to support multiple pulls and expect `"n": 1`
```
{+
    {{
        C: RUN "RETURN 1 as n" {} {"mode": "r"}
        C: PULL {"n": 1000}
        S: SUCCESS {"fields": ["n"]}
           FAILURE {"code": "Neo.ClientError.Security.AuthorizationExpired", "message": "Authorization expired"}
        S: <EXIT>
    ----
        C: RUN "RETURN 2 as n" {} {"mode": "r"}
        S: SUCCESS {"fields": ["n"]}
        {+
            C: PULL {"n": 1}
            S: RECORD [1]
               SUCCESS {"type": "r", "has_more": true}
        +}
        C: DISCARD {"n": -1}
        S: SUCCESS {}
    ----
        C: RUN "RETURN 3 as n" {} {"mode": "r"}
        C: PULL {"n": 1000}
        S: SUCCESS {"fields": ["n"]}
           RECORD [1]
           SUCCESS {"type": "r"}
    }}

+}
```
2. Configure `session1` to have `fetchSize=1`:
```
session1 = driver.session('r', fetchSize=1)
```
3. Change query in `session3`:
```
session3.run('RETURN 3 as n').next()
```

tests/stub/authorization.py
```diff
@@ -712,3 +714,114 @@ def get_vars(self, host=None):
...
+        session1.run('RETURN 2 as n').next()
+
+        try:
+            session2.run('RETURN 1 as n').next()
+        except types.DriverError as e:
+            self.assert_is_authorization_error(e)
+
+        session2.close()
+
+        session1.run('RETURN 2 as n').next()
```
injectives, line 830: With the latest changes there is a little problem here. Java pulls all records for the first query before it runs the second one, but the script expects `DISCARD` in the end.

I made it work by introducing a fork:
```
{{
    C: DISCARD {"n": -1}
    S: SUCCESS {}
----
    C: PULL {"n": -1}
    S: RECORD [1]
       SUCCESS {}
}}
```

Or we could use an entirely separate script for this test.

injectives: Thank you for adding this 👍

injectives (approved):
LGTM

robsdedude (approved):
nice

PR #108, opened 2021-05-06
Removing extra call for StartTest

bigmontz:


robsdedude:
@bigmontz that was there intentional. That call sent the new test ids while the second one sends the translated id. This would make it easier to transition between new and old ids.

bigmontz:
Sorry @robsdedude, 

PR reverting the changes:
https://github.com/neo4j-drivers/testkit/pull/119

PR #111, opened 2021-05-10
Enabling testkit to build the driver and run only one test.

bigmontz:
This option is especially useful during the development of new tests when it's used in combination with `run_all.py` enabling to run one specific test against all known drivers.

### DIFF
--- a/main.py
+++ b/main.py
@@ -37,6 +37,7 @@ test_flags = {
     "UNIT_TESTS": True,
     "INTEGRATION_TESTS": True,
     "STUB_TESTS": True,
+    "RUN_SELECTED_TESTS": False,
     "STRESS_TESTS": True,
     "TLS_TESTS": True
 }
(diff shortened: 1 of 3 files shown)


main.py
```diff
@@ -117,11 +118,17 @@ def initialise_configurations():
...
+            key
+            for (key, value) in test_flags.items()
+            if value
+        ]
 
     for item in test_flags:
         if item not in requested_list:
             test_flags[item] = False
+        else:
+            test_flags[item] = True
```
robsdedude: For a falsy `request_list` the loop will just set each entry of `test_flags` to what it already is.

robsdedude (approved):
Minor optimization possible as commented.
Didn't try it out myself but looks good code wise.

Seems like a very handy feature to have :+1:

main.py
bigmontz: I've done a bit better. 

If no request list is provider, I didn't loop through the test_flags and use its default value. Otherwise, i've loop the test_flags setting the item values equals true if it is in the requested_list, false otherwise.

PR #112, opened 2021-05-10
1st round of migrating stub tests from Python driver

robsdedude:


nutkit/protocol/feature.py
```diff
@@ -5,4 +5,21 @@
 
 
 class Feature(Enum):
+    # === OPTIMIZATIONS ===
+    # On receiving Neo.ClientError.Security.AuthorizationExpired, the driver
+    # shouldn't reuse any open connections for anything other than finishing
+    # a started job. All other connections should be re-established before
+    # running the next job with them.
     AUTHORIZATION_EXPIRED_TREATMENT = 'AuthorizationExpiredTreatment'
```
bigmontz: Since It's also an optimisation, it should have the same name scheme. We could rename the constant to `OPT_ AUTHORIZATION_EXPIRED_TREATMENT`. We could change the string itself in a next PR together with the changes on the drivers.

tests/stub/iteration/scripts/v4x0/tx_pull_1_nested.script
```diff
@@ -17,14 +17,14 @@ S: SUCCESS {"fields": ["x"], "qid": 2}
    SUCCESS {"has_more": true}
 C: PULL {"n": 1, "qid": 2}
 S: RECORD ["2_2"]
-   SUCCESS {"has_more": false}
+   SUCCESS {}
```
bigmontz, line 20: Why remove the `"has_more": false`?

tests/stub/optimizations/test_optimizations.py
```diff
@@ -0,0 +1,147 @@
...
+            driver.close()
+            self._server.done()
+
+        for mode in ("read", "write"):
+            for use_tx in (True, False):
+                with self.subTest(mode + ("_tx" if use_tx else "")):
+                    test()
+                self._server.reset()
+
+    def double_read(self, mode, new_session, use_tx, check_single_connection,
```
bigmontz, line 54: This method could be moved to the end of the file to keep the tests methods on the top. It will make it easier to navigate through the file.

bigmontz (approved):
Just a few minor comments. 

💯

tests/stub/iteration/scripts/v4x0/tx_pull_1_nested.script
robsdedude: This is what the real server does. I feel like if we want to test that the driver can handle an explicit `"has_more": false` we should probably have a separate test for it.

tests/stub/optimizations/test_optimizations.py
robsdedude: Most other test files have the common code above the test methods. I don't have a strong opinion about this, so I'll go with consistency.

robsdedude:
[Python - Integration test audit](https://trello.com/c/ekxyHUjl/33-python-integration-test-audit)

PR #114, opened 2021-05-10
Add container deletion before creation

injectives:
This is just to make sure the build does not fail if there is such container already (possibly from a previous build).

### DIFF
--- a/docker.py
+++ b/docker.py
@@ -39,8 +39,10 @@ class Container:
         del _running[self.name]
 
 
-def create(image, name, command=None, mount_map=None, host_map=None, port_map=None,
-           env_map=None, working_folder=None, network=None, aliases=None):
+def create_or_replace(image, name, command=None, mount_map=None, host_map=None,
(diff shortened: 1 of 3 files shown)


robsdedude:
A problem I faced myself often enough :+1:

docker.py
```diff
@@ -39,8 +39,10 @@ def rm(self):
         del _running[self.name]
 
 
-def create(image, name, command=None, mount_map=None, host_map=None, port_map=None,
-           env_map=None, working_folder=None, network=None, aliases=None):
+def createOrReplace(image, name, command=None, mount_map=None, host_map=None,
```
robsdedude: `create_or_replace` would be the pythonic way to go here.

robsdedude (changes requested):

docker.py
injectives: fixed 👍

robsdedude (approved):
Didn't run it myself, but looks good to me.

PR #117, opened 2021-05-11
Add consumption to auth expired on pull tests

injectives:
Add consumption to auth expired on pull tests

### DIFF
--- a/tests/stub/authorization.py
+++ b/tests/stub/authorization.py
@@ -238,7 +238,8 @@ class AuthorizationTests(BaseAuthorizationTests):
 
         session = driver.session('r', database=self.get_db())
         try:
-            session.run("RETURN 1 as n")
+            result = session.run("RETURN 1 as n")
+            result.consume()
         except types.DriverError as e:
             self.assert_is_authorization_error(error=e)
         session.close()
@@ -290,7 +291,8 @@ class AuthorizationTests(BaseAuthorizationTests):
         session = driver.session('r', database=self.get_db())
         tx = session.beginTransaction()
         try:
-            tx.run("RETURN 1 as n").consume()
+            result = tx.run("RETURN 1 as n")
+            result.consume()
         except types.DriverError as e:
             self.assert_is_authorization_error(error=e)
         session.close()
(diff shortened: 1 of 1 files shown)


tests/stub/authorization.py
```diff
@@ -238,7 +238,7 @@ def test_should_fail_with_auth_expired_on_pull_using_session_run(
 
         session = driver.session('r', database=self.get_db())
         try:
-            session.run("RETURN 1 as n")
+            session.run("RETURN 1 as n").consume()
```
robsdedude: I'd suggest to split the two calls into two statements:
```
res = session.run("RETURN 1 as n")
res.consume()
```

This might make it easier to debug which call caused an error, should one be raised.

injectives: sure, I have updated the PR

AndyHeap-NeoTech (approved):
This looks good to me.

tests/stub/authorization.py
```diff
@@ -290,7 +291,8 @@ def test_should_fail_with_auth_expired_on_run_using_tx_run(self):
         session = driver.session('r', database=self.get_db())
         tx = session.beginTransaction()
         try:
-            tx.run("RETURN 1 as n").consume()
+            result = tx.run("RETURN 1 as n")
+            result.consume()
```
robsdedude, line 295: Arguably out of scope of these tests, but do (and do we want to allow it) some drivers  keep "RUN" in their buffer until the result is consumed?

tests/stub/authorization.py
```diff
@@ -339,7 +342,8 @@ def test_should_fail_with_auth_expired_on_commit_using_tx_run(self):
 
         session = driver.session('r', database=self.get_db())
         tx = session.beginTransaction()
-        tx.run("RETURN 1 as n").consume()
+        result = tx.run("RETURN 1 as n")
+        result.consume()
```
robsdedude, line 346: Consume shouldn't be needed here if commit causes the error, should it?

tests/stub/authorization.py
injectives: I have only updated this to be consistent with your previous change suggestion. It should be identical to what it was before.
In general, I think some drivers (java and probably dotnet) require result to be consumed to display the error.

tests/stub/authorization.py
injectives: I have only updated this to be consistent with your previous change suggestion. It should be identical to what it was before.
Potentially it is not needed. But I would suggest looking into it in a separate PR if required.

robsdedude (approved):

PR #118, opened 2021-05-11
Fix one test and new message

robsdedude:
* Fix a tests that was broken for JS
* `GetFeatures` message would cause the Java driver to fail the first test because it closes the connection on receiving an unknown message.

### DIFF
--- a/tests/shared.py
+++ b/tests/shared.py
@@ -69,7 +69,11 @@ class MemoizedSupplier:
 
 @MemoizedSupplier
 def get_driver_features(backend):
-
+    # TODO: remove this as soon as java supports GetFeatures
+    if get_driver_name() in ["java"]:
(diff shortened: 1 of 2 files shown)


tests/shared.py
```diff
@@ -69,7 +69,11 @@ def __call__(self, *args, **kwargs):
 
 @MemoizedSupplier
 def get_driver_features(backend):
-
+    # TODO: remove this as soon as java supports GetFeatures
+    if get_driver_name() in ["java"]:
```
bigmontz, line 73: See https://github.com/neo4j/neo4j-java-driver/pull/889#issuecomment-838462619

robsdedude: But it's not merged yet. So all pipelines until then will be blocked :/

robsdedude:
Closed in favor of #117

PR #121, opened 2021-05-11
Remove redundant consume and add a note regarding failures on run

injectives:
Remove redundant consume and add a note regarding failures on run in authorization expired tests.


robsdedude (approved):
I'm not sure anymore if the consume that received the comment should actually be removed or even mandatory. It definitely should be uniform across the drivers (i.e., only need one call to break: `run` *or* `consume`). I've added a [card](https://trello.com/c/2pVezWj0/3588) so we can decide that as a team later.

injectives:
There was a conversation on this previously and I left a comment with more details.

PR #122, opened 2021-05-14
Set the correct scheme to connect to the neo4j server

bigmontz:
The driver was always using the `bolt` as scheme for the `uri` during the testkit/neo4j tests. This behaviour can lead to error, especially during the tests which performs writes.


robsdedude (approved):
LGTM

PR #124, opened 2021-05-17
Fix accepted_messages in stub server

robsdedude:
`.accepted_messages` would not always be correct on all scrip blocks. This could lead to `ScriptDeviation` exceptions to not list all accepted messages or, in some edge cases, raise an `IndexError`.


robsdedude:
Pipelines are green. The publisher just failed to update the state on this PR

bigmontz (approved):
🉑

PR #125, opened 2021-05-17
Standardise methods names at `datatypes.py`

bigmontz:


robsdedude (approved):
:+1: I suggest to wait for the pipelines to pass though, because some drivers might skip those tests and this PR changes the test IDs.

PR #128, opened 2021-05-19
2nd round of python stub test migration

robsdedude:


robsdedude:
[Python - Integration test audit](https://trello.com/c/ekxyHUjl/33-python-integration-test-audit)

tests/shared.py
```diff
@@ -136,6 +132,19 @@ def setUp(self):
                             "received {}: {}".format(type(response),
                                                      response))
 
+    def driver_missing_features(self, *features):
+        needed = set(map(lambda f: f.value, features))
+        supported = self._driver_features
+        return needed - supported
+
+    def driver_supports_features(self, *features):
```
injectives, line 140: Sorry, is this actually used somewhere?

robsdedude: It will be used in my next round of test migration. Even though, each test should only test one thing, I decided that because spinning up a stub server is relatively expensive, I would add additional assertions that only make sense for driver supporting certain features.

I did it in this PR, because I thought I would need it but didn't actually end up using it. But I still think it's a neat feature to have.

tests/stub/disconnects/scripts/return_1.script
```diff
@@ -0,0 +1,17 @@
...
+!: AUTO HELLO
+
+{*
+    C: RESET
+    S: SUCCESS {}
+*}
+C: RUN "*" "*" "*"
+S: SUCCESS {"fields": ["n.name"]}
+C: PULL {"n": "*"}
+S: SUCCESS {"type": "w"}
```
injectives: Am I missing something or does this script return nothing? I was expecting one record based on its name :)

tests/stub/iteration/test_iteration_session_run.py
```diff
@@ -61,7 +61,40 @@ def test_error(self):
...
+
     # Support -1, not batched at all for BOLTv3
     def test_all_v3(self):
         self._run(-1, "pull_all.script", ["1", "2", "3", "4", "5", "6"],
                   protocol_version="v3")
+
+    def test_discards_on_session_close(self):
+        # TODO remove this block once all languages work
+        if get_driver_name() in ['java']:
+            self.skipTest("Eagerly pulls more results")
```
injectives: I had a quick look at java.
The reason this test may fail is because it implements a concept of `low and high watermarks`.
In Bolt 4+ the watermarks are defined as follows:
```
highRecordWatermark = (long) (fetchSize * 0.7); // would be 1 when fetchSize is 2
lowRecordWatermark = (long) (fetchSize * 0.3);  // would be 0 when fetchSize is 2
```

The `lowRecordWatermark` is checked when record is returned to the client code:
```
if ( records.size() <= lowRecordWatermark ) {
    .........
    request( fetchSize ); // while in streaming state, this actually is used to calculate how many records to fetch when Success arrives
    .........
}
```

Given the test code, the following happens:
1. Query is submitted to the server
2. First record arrives
3. Because the client is waiting for the first record, its dequeuing begins
4. Low watermark is checked against the internal buffer, which is empty at that point
5. Future fetch is noted internally
6. The first record is returned to the client (client gets unblocked)
7. The internals process the next server responses (RECORD AND SUCCESS)

From this point, depending on timing, at least the following 2 things may happen:
1. If SUCCESS arrives before the client closes the session, the underlying logic will submit another PULL due to the low watermark signal. To account for this, the stub server script may need to support an additional optional PULL.
2. If SUCCESS arrives after the client closes the session, the DISCARD will be sent. This can be reproduced by putting sleep between first and second records in the stub server script.

If this is just one test, we may want to support the additional optional PULL. Otherwise, we may actually want to bring the `watermark feature` awareness to testkit.

tests/shared.py
injectives: Sure, as long as you are confident that you will use it in the next PR and it will not remain unused.

tests/stub/versions/test_versions.py
```diff
@@ -20,21 +25,73 @@ def tearDown(self):
...
+            driver.close()
+            self._server.reset()
+
+    def _run(self, version, server_agent=None, check_version=False,
+             rejected_agent=False, check_server_address=False):
+        """
+            :param version: e.g. "4x3" or "3"
+            :type version: string
+            :param server_agent: Set a specific server agent and check that the
+                                 summery list the correct one.
```
injectives: Potential typos :)

tests/stub/versions/test_versions.py
```diff
@@ -20,21 +25,73 @@ def tearDown(self):
...
+            :type server_agent: str, optional
+            :param check_version: Check if the summery contains the correct bolt
```
injectives: I assume this meant to be `summary`.

tests/stub/versions/test_versions.py
```diff
@@ -20,21 +25,73 @@ def tearDown(self):
...
+                                  version
+            :type check_version: bool
+            :param rejected_agent:
+                True: Expect the driver to fail by rejecting the server agent
+                False: Expect the driver not to fail
+            :type rejected_agent: bool
+            :param check_server_address: Check if the summery contains the right
```
injectives: I assume this meant to be `summary`.

tests/stub/versions/test_versions.py
```diff
@@ -20,21 +25,73 @@ def tearDown(self):
...
+                        self.assertEqual(summary.server_info.address,
+                                         self._server.address)
+                else:
+                    # Otherwise the script will not fail when the protocol is
+                    # not present (on backends where run is lazily evaluated)
+                    result.next()
+            except DriverError:
+                if not rejected_agent:
+                    raise
+                # TODO: check exception further
```
injectives, line 90: Are you sure you want to keep this comment?

tests/stub/versions/test_versions.py
```diff
@@ -52,3 +109,59 @@ def test_supports_bolt_4x3(self):
...
+                ("FooBar/4.3.0", True),
+            ):
+                if get_driver_name() in ["dotnet"]:
+                    # skip subtest: Doesn't reject server's agent sting,
+                    #               compiles server agent from bolt version
+                    #               potentially more differences
+                    continue
+                if reject and get_driver_name() in ["javascript", "go", "java"]:
+                    # skip subtest: Does not reject server's agent string
+                    continue
```
injectives, line 136: Does this result in test being displayed as skipped in build output?

tests/stub/versions/test_versions.py
robsdedude: I think, once we agree on a common exception hierarchy/naming and add testkit messages for them, this test should be extended to check for the correct error to be thrown.

tests/stub/versions/test_versions.py
robsdedude: Sadly, no. Unittest's sub-tests don't support skipping. You either skip the whole test (all sub-tests) or none. Sub-tests are also not counted as own tests, iirc. However, this if-branch will leave some sub-tests to be run for those three drivers.

tests/stub/disconnects/scripts/return_1.script
robsdedude: No need to return records for this test. But the script's name is misleading indeed.

tests/stub/iteration/test_iteration_session_run.py
robsdedude: Solved it by allowing more `PULL`s in the script and asserting only one `PULL` happens for all but the JAVA driver. A long term solution should be found (see Trello).

tests/stub/versions/test_versions.py
robsdedude: Add TODO comment

injectives (approved):
LGTM

PR #129, opened 2021-05-20
Update stub server log output to contain in and out port

robsdedude:
This helps to debug scripts that allow concurrent connections


bigmontz (approved):
🦞

PR #134, opened 2021-05-25
Fix stub server not handling non-determinism right

robsdedude:
Non-deterministic blocks inside alternative or parallel blocks could cause the
stub server to crash.


bigmontz (approved):
💯

PR #135, opened 2021-05-27
Add OPT_AUTHORIZATION_EXPIRED_TREATMENT to all authorization expired tests

injectives:
Add OPT_AUTHORIZATION_EXPIRED_TREATMENT to all authorization expired tests

### DIFF
--- a/tests/stub/authorization.py
+++ b/tests/stub/authorization.py
@@ -222,6 +222,7 @@ class AuthorizationTests(BaseAuthorizationTests):
                 server.start(script=new_script,
                              vars=self.get_vars())
 
+    @driver_feature(types.Feature.OPT_AUTHORIZATION_EXPIRED_TREATMENT)
     def test_should_fail_with_auth_expired_on_pull_using_session_run(
             self):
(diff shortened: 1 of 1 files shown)


robsdedude (approved):
My thoughts on this:  
 - In the long run, we definitely want to add this feature flags to all tests where applicable. I just didn't go for it right now because it would disable quite a few tests for drivers that haven't implemented feature messages yet.
 - Optimally, the 4.2 driver shouldn't be tested with testkit 4.3.

injectives:
> My thoughts on this:
> 
> * In the long run, we definitely want to add this feature flags to all tests where applicable. I just didn't go for it right now because it would disable quite a few tests for drivers that haven't implemented feature messages yet.
> * Optimally, the 4.2 driver shouldn't be tested with testkit 4.3.

In general, this makes sense to me.
On the first point, I believe it is reasonable approach in this particular case as the feature flag is already used for these tests.
On the second point, I think this will become less of a problem as we transition onto the 4.3 branches. There was a desire to use testkit tests and benefit from the latest improvements we have been introducing.

PR #138, opened 2021-05-28
Fix stub-server hanging

robsdedude:


boltstub/__init__.py
```diff
@@ -111,11 +112,11 @@ def setup(self):
                          self.client_address, self.server_address)
 
             def handle(self):
+                with service.actors_lock:
```
bigmontz, line 115: Why did it moved out of the try catch?

robsdedude: The part I excluded from the `try` should never fail. It's just initializing the Script object without any interactions with the driver. And if it ever fails, I'm happy with the whole stub service crashing. This way it's easier to debug.

I'm not sure https://github.com/neo4j-drivers/testkit/pull/138/files/afdf93b8e839510b420b357dd207123fcdb29ede#diff-5a7c05dad1b40630c3253aa31e8e7bd6eff29623c43576574b4fc747fa42f07bR128 is a good idea anyway.

PR #142, opened 2021-05-31
Increase stubserver stability

robsdedude:
Under certain circumstances, the stub server could crash when it tried to obtain
a connection's port for logging purposes after the connection had been closed
already.


bigmontz (approved):
🏄🏼

PR #143, opened 2021-06-07
Jolt support

robsdedude:
Make the stub server understand JOLT format for non-JSON-native types in client and server lines.

Note that the implementation diverges from the JOLT specs in that it adds a simple form for dicts whenever this is unambiguous to enable backwards compatibility with existing stubscripts.


boltstub/packstream.py
```diff
@@ -44,22 +46,129 @@
...
+                    or not all(isinstance(k, str) for k in fields[2].keys())):
+                raise ValueError(
+                    "Invalid UnboundRelationship struct received %r" % self
+                )
+
+        def verify_path():
+            if (len(fields) != 3
+                    or not isinstance(fields[0], list)
+                    or not all(isinstance(n, Structure)
+                               and n.tag == b"\x4E"
```
bigmontz, line 98: Maybe we could have this magic binaries in some "constant", it could be a bit difficult to keep track of this values sometimes.

boltstub/simple_jolt/transformers.py
```diff
@@ -0,0 +1,603 @@
...
+                                               human_readable=human_readable)
+            except NoFullRepresentation:
+                return transformer.encode_simple(value_, transform,
+                                                 human_readable=human_readable)
+        return value_
+    return transform(value)
+
+
+__all__ = [
+    # JoltNullTransformer,
```
bigmontz, line 587: Why do we have this lines commented?

bigmontz (approved):
🥇 

(Just two small comments)

PR #145, opened 2021-06-09
Restructure server side routing tests

robsdedude:


tests/stub/server_side_routing/test_server_side_routing.py
```diff
@@ -70,14 +50,22 @@ def test_direct_connection_without_url_params(self):
...
-        except Exception:
+        except types.DriverError:
             pass
+        except types.BackendError:
+            if get_driver_name() in ["javascript"]:
+                # TODO: this shouldn't be communicated as backend error
+                return
+        except Exception as e:
+            if get_driver_name() in ["java"]:
+                if "TestkitErrorResponse" in str(e):
```
injectives, line 64: This is a general backend error. The underlying cause is:
```
java.lang.IllegalArgumentException: Routing parameters are not supported with scheme 'bolt'. Given URI: 'bolt://localhost:9001?region=china&policy=my_policy'
```

robsdedude: Yeah, but that message type the Java back end sends, does not exist. Only `DriverError` and `BackendError` are defined in the testkit protocol (technically also `FrontendError` but that one is irrelevant here). In this case I'd expect a `DriverError` to be returned as it's the driver code that complains about what's going on here.

injectives: I will note this in a card and should be able to fix it.

tests/stub/server_side_routing/test_server_side_routing.py
```diff
@@ -70,14 +50,22 @@ def test_direct_connection_without_url_params(self):
...
-                           })
+        self._start_server()
         try:
             driver = Driver(self._backend, uri, self._auth, self._userAgent)
-        except Exception:
+        except types.DriverError:
             pass
+        except types.BackendError:
+            if get_driver_name() in ["javascript"]:
+                # TODO: this shouldn't be communicated as backend error
```
bigmontz, line 60: Sure it's an issue with the backend, it's not capturing correctly the exception.

tests/stub/server_side_routing/test_server_side_routing.py
```diff
@@ -70,14 +50,22 @@ def test_direct_connection_without_url_params(self):
...
+                return
+        except Exception as e:
+            if get_driver_name() in ["java"]:
+                if "TestkitErrorResponse" in str(e):
+                    # TODO: sends unknown protocol message on failure
+                    return
+            raise
+
         else:
             # Python driver
```
injectives, line 70: I assume we would eventually want to align the behaviour?

robsdedude: Yes. I also don't like that the Python driver needs an extra `else` clause here. But that's out of scope of this PR. I just wanted to move the tests into their own package like all the other tests. While I was at it, I cleaned up the code a bit. Like here, for example, `except:` or `except Exception:` is rarely a good idea in Python. It also catches `KeyboardInterrup` (and other exceptions), which you most likely want to let bubble up.

bigmontz (approved):
🍎

tests/stub/server_side_routing/test_server_side_routing.py
injectives: No problem, just wanted to ask if you had it on your list 👍

injectives (approved):
LGTM

PR #149, opened 2021-06-14
Fix docker complaining about relative paths

robsdedude:
Specifying a relative path (e.g., `TEST_DRIVER_REPO=../neo4j-python-driver`) would cause docker to complain:  
`Error response from daemon: create ../neo4j-python-driver: "../neo4j-python-driver" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path
`


michael-simons (approved):
I like that change a lot :)

PR #151, opened 2021-06-15
Fix server address checks in summary when run inside docker

robsdedude:
Inside docker, the server addresses will look like "runner:1234". We expect the driver to return the resolved IP address in the summary.

### DIFF
--- a/tests/stub/shared.py
+++ b/tests/stub/shared.py
@@ -12,6 +12,7 @@ from queue import (
 )
 import re
 import signal
+import socket
 import subprocess
 import sys
 import tempfile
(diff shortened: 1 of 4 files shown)


tests/stub/shared.py
```diff
@@ -382,3 +383,20 @@ def pick_address(adapter_):
             ips.append(address)
 
     return ips
+
+
+def dns_resolve(addr):
+    return socket.gethostbyname_ex(addr)[2]
```
bigmontz: Maybe:
```
_, _, ips = socket.gethostbyname_ex(addr)
return ips
```

Or something similar. It makes easier to realize what's the method is returning.

bigmontz (approved):
🕷️

PR #156, opened 2021-06-29
1st round of migrating Python integration test

robsdedude:


nutkit/protocol/feature.py
```diff
@@ -14,12 +14,25 @@ class Feature(Enum):
...
     # The driver doesn't wait for a SUCCESS after calling RUN but pipelines a
-    # PULL right afterwards and consumes two message after that. This saves a
+    # PULL right afterwards and consumes two messages after that. This saves a
     # full round-trip.
     OPT_PULL_PIPELINING = "Optimization:PullPipelining"
+
+    # Temporary driver feature that will be removed when all official drivers
+    # have been unified in their behaviour of when they return a Result object.
+    # We aim for drivers to not providing a Result until the server replied with
+    # SUCCESS so that the result keys are already know when and attached to the
```
injectives: (Optional) I believe it should be `known` and the following `when` may be redundant. You may want to review this paragraph.

nutkit/protocol/responses.py
```diff
@@ -196,19 +196,105 @@ def __eq__(self, _):
...
+                else:
+                    data[field] = None
+        # ----------------------------------------------------------------------
         self.server_info = ServerInfo(**data["serverInfo"])
+        self.counters = Counters(**data["counters"])
+        self.database = data["database"]
+        self.notifications = data["notifications"]
+        self.plan = data["plan"]
+        self.profile = data["profile"]
+        self.parameters = data["parameters"]
```
injectives: I haven't actually seen this in java implementation.....

tests/neo4j/suites.py
```diff
@@ -90,6 +102,9 @@
         print("Unknown suite name: " + name)
         sys.exit(-1)
 
+    import os
+    os.environ["env_neo4j_version"] = os.environ.get("env_neo4j_version", name)
```
injectives, line 106: Out of interest, how is this used?

tests/neo4j/test_direct_driver.py
```diff
@@ -0,0 +1,163 @@
...
+
+        host, port = get_neo4j_host_and_port()
+        host = dns_resolve_single(host)
+        resolved_addresses = []
+
+        def my_resolver(socket_address):
+            resolved_addresses.append(socket_address)
+            return (
+                # should be rejected as unable to connect
+                "99.99.99.99:%d" % port,
```
injectives: It could also be some ip from the loopback range (`127.0.0.0/8`).

tests/neo4j/test_summary.py
```diff
@@ -0,0 +1,234 @@
...
+        self.assertEqual(summary.query, "RETURN $number AS x")
+        self.assertEqual(summary.parameters, params)
+
+        self.assertIn(summary.query_type, ("r", "w", "rw", "s"))
+
+        self._assert_counters(summary)
+
+    @driver_feature(types.Feature.TMP_RESULT_KEYS,
+                    types.Feature.TMP_FULL_SUMMARY)
+    def test_summary_counters_case_2(self):
```
injectives, line 124: I assume it is a ported version of an existing test. In the future, it may be worth splitting it up.

robsdedude: Yes, the tests weren't particularly well named or structured. Maybe I should just do the splitting up right now, before actually merging in this tech debt.

tests/neo4j/suites.py
robsdedude: This, together with [this](https://github.com/neo4j-drivers/testkit/pull/156/files#diff-81d6da2c1420e0c0ab1903e78fec06d8ab0d78483f4e47abe41c82cd2aa4ef36R65-R92), [this](https://github.com/neo4j-drivers/testkit/pull/156/files#diff-34cde5e49b54d9b67952c80d3580b6c421ebcf16097996e215aef54e44f78562R77-R79), and [that](https://github.com/neo4j-drivers/testkit/pull/156/files#diff-34cde5e49b54d9b67952c80d3580b6c421ebcf16097996e215aef54e44f78562R102-R104) is used to know which features can be used in tests (e.g. multi-database).

nutkit/protocol/responses.py
robsdedude: I saw this in the docs `ResultSummary.query().parameters()`. Note that the query parameters are nested inside a query object for some drivers, while others expose them at summary level.

robsdedude: Actually, after checking, I found that only Python goes without the nesting. Maybe something to consider for 5.0. Do you think I should adjust the testkit protocol here to reflect how most drivers do it, @injectives?

injectives: I believe we have 2 options:
1. Update to work as a majority for now.
2. Discuss unification and work from there. We could update drivers and retain the previous methods in deprecated state until 5.0.

robsdedude: I chose both. 1. and 2. We can't unify the surface until 5.0 either way.

injectives (approved):
👍

PR #159, opened 2021-07-08
Make sure stub server tests protocol version negotiation more strictly

robsdedude:


robsdedude:
Python is expected to fail until https://github.com/neo4j/neo4j-python-driver/pull/559 is merged which in turn depends on this one.

PR #162, opened 2021-07-14
2nd round of migrating Python integration test

robsdedude:


tests/neo4j/sessionrun.py
```diff
@@ -207,6 +207,54 @@ def test_iteration_larger_than_fetch_size(self):
             rec = result.next()
             self.assertEqual(rec, exp)
 
+    def test_partial_iteration(self):
```
bigmontz, line 210: This test seems to be testing too many scenarios, maybe it could be split in two. 

One with `# Verifies that not consuming all records works` and `# not consumed all records & starting a new session`. 

The other one with `# not consumed all records & starting a new session` and `# not consumed all records & reusing the previous session`. 

But I'm not sure if it's the best way to split it. My concern here is not knowing just by look the test is failing what's wrong.

tests/neo4j/sessionrun.py
```diff
@@ -207,6 +207,54 @@ def test_iteration_larger_than_fetch_size(self):
...
+            self.assertEqual(records,
+                             [[i] for i in map(types.CypherInt, range(5, 9))])
+            summary = result.consume()
+            self.assertIsInstance(summary, types.Summary)
+            self._session1.close()
+            self._session1 = None
+
+        for consume in (True, False):
+            with self.subTest("consume" if consume else "no consume"):
+                _test()
```
bigmontz, line 256: The readability of the code could improve with the `consume` flag sent by param to the `_test` function, instead of getting it from the context.

tests/neo4j/txrun.py
```diff
@@ -118,18 +122,143 @@ def test_should_fail_run_in_a_rollbacked_tx(self):
...
+        if self.driver_supports_features(types.Feature.TMP_RESULT_KEYS):
+            self.assertEqual(result.keys(), ["x"])
+        values = []
+        for _ in range(2):
+            record = result.next()
+            self.assertIsInstance(record, types.Record)
+            values.append(record.values[0])
+        tx.commit()
+
+        # TODO: what should happen here?
```
bigmontz, line 166: Which piece of code does this todo refer?

tests/neo4j/txrun.py
```diff
@@ -118,18 +122,143 @@ def test_should_fail_run_in_a_rollbacked_tx(self):
...
+            else:
+                values1 = list(map(lambda rec: rec.values[0], result1))
+                values2 = list(map(lambda rec: rec.values[0], result2))
+            tx.commit()
+            self.assertEqual(values1, list(map(types.CypherInt, (1, 2, 3, 4))))
+            self.assertEqual(values2, list(map(types.CypherInt, (5, 6, 7, 8))))
+            self._session.close()
+            self._session = None
+
+        for invert_fetching in (True, False):
```
bigmontz, line 193: Same observation about the implicitly variable.

tests/neo4j/txrun.py
```diff
@@ -118,18 +122,143 @@ def test_should_fail_run_in_a_rollbacked_tx(self):
...
+                result2 = tx.run("UNWIND [5,6,7,8] AS y RETURN y")
+            if self.driver_supports_features(types.Feature.TMP_RESULT_KEYS):
+                self.assertEqual(result2.keys(), ["y"])
+            values2 = list(map(lambda rec: rec.values[0], result2))
+
+            self.assertEqual(values2, list(map(types.CypherInt, (5, 6, 7, 8))))
+            tx.commit()
+            self.assertIsInstance(result2.next(), types.NullRecord)
+            self.assertEqual(values1, [types.CypherInt(1)])
+            # TODO: what should result1.next() result in?
```
bigmontz, line 220: Or It should return all the already buffered records until and then fails when trying fetching more results or it should fail immediately.

tests/neo4j/txrun.py
```diff
@@ -118,18 +122,143 @@ def test_should_fail_run_in_a_rollbacked_tx(self):
...
+                tx.commit()
+            else:
+                tx.rollback()
+
+            res = self._session.run("MATCH (a:Thing {uuid:$uuid}) RETURN a",
+                                    params={"uuid": types.CypherString(uuid_)})
+            self.assertEqual(len(list(res)), commit)
+
+        for commit in (True, False):
+            with self.subTest("commit" if commit else "rollback"):
```
bigmontz, line 263: Same observation about the implicitly variable.

bigmontz (approved):
🥇

tests/neo4j/sessionrun.py
robsdedude: This is a good point and I actually thought about splitting it. The thing is, that these integration tests are rather costly in terms of runtime. And if I split them, I'll have to send 4 instead of 3 queries to the server + whatever fluff is needed around it (e.g., TestKit protocol communication, etc).

tests/neo4j/sessionrun.py
robsdedude: Yes and no. I'd rather avoid shadowing outer variables. So I'd have to call it something else inside the function. Which do you prefer?
 1. ```python
    def foo(bar_):
        if bar_:
           pass  # do something
   
    bar = True  # or whatever
    foo(bar)
    ```
 2. ```python
    def foo():
        if bar:
            pass  # do something
    
    bar = True  # or whatever
    foo()
    ```

tests/neo4j/txrun.py
robsdedude: There should be an assertion, but all official drivers behave differently here. So before we add something here, we need to decide what that something should be. Here is the respective [Trello card](https://trello.com/c/F1JAhHbz/81-unmanaged-tx-result-scope).

tests/neo4j/txrun.py
robsdedude: True. Keep that in mind for the next unification meeting, please!

tests/neo4j/sessionrun.py
bigmontz: Maybe:

```python
def _test(consume):
    if consume:
        pass # do something

 
with self.subTest("consume"):
    _test(consume=True)

with self.subTest("no consume"):
    _test(consume=False)
```

Or use the _consume in the `for _consume in (True, False):`

But I agree it's a tough decision.

robsdedude: I should just transition TestKit to pytest asap. It would make all these problems go away as it has proper test parametrization support. Anyway, until then, I'd leave it as is unless you definitely want me to change it. Regarding your 1st suggestions: looks the cleanest, admittedly.,but it falls short if you want to parametrize more than one thing.

PR #187, opened 2021-07-29
Add connection hint test for timed out transaction

injectives:
Cherry-pick of: #185.

### DIFF
--- a/tests/stub/configuration_hints/test_connection_recv_timeout_seconds.py
+++ b/tests/stub/configuration_hints/test_connection_recv_timeout_seconds.py
@@ -36,13 +36,19 @@ class TestDirectConnectionRecvTimeout(TestkitTestCase):
     def _start_server(self, script):
         self._server.start(path=self.script_path(script))
 
-    def _assert_is_timout_exception(self, e):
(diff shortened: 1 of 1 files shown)


bigmontz (approved):
🍒

tests/stub/configuration_hints/test_connection_recv_timeout_seconds.py
```diff
@@ -90,12 +96,42 @@ def test_timeout_unmanaged_tx(self):
...
+            # has async iterator api
+            if get_driver_name() in ['javascript']:
+                result.next()
+
+        try:
+            result = tx.run("in time")
+            if get_driver_name() in ['javascript']:
+                result.next()
+        except types.DriverError as e:
+            self._assert_is_client_exception(e)
```
robsdedude, line 122: This does not assert that an exception is raised. If no error occurs, the test still passes.

```python
with self.assertRaises(types.DriverError) as e:
    run_stuff(something)
self._assert_is_client_exception(e.exception)
```

injectives: Good point, I will update it. Thank you 👍

PR #190, opened 2021-08-03
Improve OPT_MINIMAL_RESETS test coverage

robsdedude:


tests/stub/disconnects/test_disconnects.py
```diff
@@ -250,6 +250,9 @@ def test_disconnect_session_on_tx_commit(self):
     # FIXME: This test doesn't really fit here. It tests FAILURE handling, not
     #        handling sudden loss of connectivity.
     def test_fail_on_reset(self):
+        if self.driver_supports_features(types.Feature.OPT_MINIMAL_RESETS):
+            self.skipTest("The failing RESET never gets issued by an optimized "
+                          "driver.")
```
bigmontz, line 255: 💯

tests/stub/optimizations/scripts/v4x3/pull_pipeline.script
```diff
@@ -4,7 +4,7 @@
 
 {*
     C: RESET
-    S: SUCCESS
+    S: SUCCESS {}
```
bigmontz, line 7: Is it just for clarity?

tests/stub/optimizations/test_optimizations.py
```diff
@@ -25,7 +27,7 @@ def tearDown(self):
     def test_pull_pipelining(self):
         def test():
             script = "pull_pipeline{}.script".format("_tx" if use_tx else "")
-            self._server.start(path=self.script_path(script),
+            self._server.start(path=self.script_path("v4x3", script),
```
bigmontz, line 30: Why does the pull pipelining is now only available for the 4.3 scenario?

tests/stub/optimizations/test_optimizations.py
```diff
@@ -119,20 +150,34 @@ def test_reuses_connection(self):
...
+                        with self.subTest(mode
+                                          + "_" + version
+                                          + (("_tx_" + use_tx) if use_tx
+                                             else "")
+                                          + ("_one_shot_session" if new_session
+                                             else"_reuse_session")
+                                          + ("_discard" if consume
+                                             else "_pull")):
+                            self.double_read(mode, new_session, use_tx, False,
+                                             False, True, version=version,
```
bigmontz: The unnamed booleans params doesn't give much context what's happening here.

bigmontz (approved):
🗡️

tests/stub/optimizations/scripts/v4x3/pull_pipeline.script
robsdedude: Nope. It could make drivers fail that expect every `SUCCESS` message to have one field `metadata::Dictionary` which is what the docs say. So the driver would be right in doing so.

tests/stub/optimizations/test_optimizations.py
robsdedude: It was like this previously. I just moved files.

tests/stub/optimizations/test_optimizations.py
robsdedude: That's true. Probably worth refactoring this. :hammer_and_wrench:

PR #193, opened 2021-08-03
Avoid using !:AUTO GOODBYE

robsdedude:
This makes sure that server.done() actually asserts that the script has been
fully played. With auto GOODBYE the driver can reach the end of the script at
any point by sending a GOODBYE which is likely to happen on driver.close() which
should be called before checking server.done() to avoid race conditions in which
the server has not yet transmitted all received messages to TestKit.


tests/stub/routing/scripts/v3/router_yielding_writer1.script
```diff
@@ -9,3 +8,7 @@ C: RUN "CALL dbms.cluster.routing.getRoutingTable($context)" {"context": #ROUTIN
 S: SUCCESS {"fields": ["ttl", "servers"]}
    RECORD [1000, [{"addresses": ["#HOST#:9001"], "role":"ROUTE"}, {"addresses": ["#HOST#:9010", "#HOST#:9011"], "role":"READ"}, {"addresses": ["#HOST#:9020"], "role":"WRITE"}]]
    SUCCESS {"type": "r"}
+{?
+    C: GOODBYE
+    S: <EXIT>
+?}
```
bigmontz, line 14: It has the potential to be a macro like `!: AUTO GOODBYE` but which only adds this snippet at the and of the script

tests/stub/routing/test_routing_v4x3.py
```diff
@@ -672,7 +672,7 @@ def test_should_fail_when_writing_on_writer_that_returns_database_unavailable(se
                 **self.get_vars(),
                 "#FAILURE#":
                     '{"code": '
-                    '"Neo.ClientError.General.DatabaseUnavailable", '
+                    '"Neo.TransientError.Database.DatabaseUnavailable", '
```
bigmontz: Why did this error code change?

tests/stub/routing/test_routing_v4x3.py
```diff
@@ -684,19 +684,19 @@ def test_should_fail_when_writing_on_writer_that_returns_database_unavailable(se
         except types.DriverError as e:
             if get_driver_name() in ['python']:
                 self.assertEqual(
-                    "<class 'neo4j.exceptions.ClientError'>",
+                    "<class 'neo4j.exceptions.TransientError'>",
```
bigmontz: Why did this error code change?

tests/stub/routing/test_routing_v4x3.py
```diff
@@ -684,19 +684,19 @@ def test_should_fail_when_writing_on_writer_that_returns_database_unavailable(se
...
                     e.errorType
                 )
                 self.assertEqual(
-                    "Neo.ClientError.General.DatabaseUnavailable",
+                    "Neo.TransientError.Database.DatabaseUnavailable",
```
bigmontz: Why did this error code change?

bigmontz (changes requested):
It needs a few clarifications about the error code changes.

tests/stub/routing/scripts/v3/router_yielding_writer1.script
robsdedude: I thought about it. But I think this would make things unintuitive as the semantics of `!: AUTO <SOME_MESSAGE>` are that the driver can send that message at any time and the stubserver will respond to it accordingly. Also, what you don't see explicitly in this PR: there are some scripts where I left the `!: AUTO GOODBYE` as is because those script were meant to be be played fully or partially without asserting that the end would be reached. I thought about adding something like `!: AUTO GOODBYE END`. If I decide to do that, I'd also like to add `!: AUTO HELLO BEGIN` as currently,  `!: AUTO HELLO` would also allow the driver to send `HELLO` in the middle of the script, which makes little sense. But the rabbit-hole goes deeper... What about `RESET` many scripts use `!AUTO: RESET` which I think is a bad idea because it will not complain about drivers `RESET`in for instance in the middle of a transaction whereas this would end the transaction on a real server.

Bottom line conclusion: I'd like to get rid of all `AUTO` statements. They should all be replaced with with an optional clause. Or a command line `C: <AUTO> GOODBYE` which would allow 0..n GOODBYE messages to be received and automatically respond accordingly. So basically `!: AUTO ...` but only at the specified spots in the script. I just don't like that particular syntax.

tests/stub/routing/test_routing_v4x3.py
robsdedude: Because that's the real error code the server issues. I could and possibly should put this into a separate PR. I just dislike the overhead of PRs.

tests/stub/routing/scripts/v3/router_yielding_writer1.script
robsdedude: What do you think about `?: <SOME_MESSAGE>` for this syntax?

bigmontz (approved):

bigmontz:
👋🏼

tests/stub/routing/scripts/v3/router_yielding_writer1.script
bigmontz: I like this clause, it reminds the `{?` and the `!:`, so we could guess without previous knowledge about what this thing will actually do.

PR #194, opened 2021-08-04
Adding explicit auto lines to stubscript

robsdedude:
`!: AUTO <SOME_COMMAND>` has the big drawback that it allows driver to send
the specified message at any given time. This can lead to tests passing that
actually shouldn't. The biggest pitfall is `!: AUTO GOODBYE` which more or less
renders `server.done()` useless as it allows the driver at any point in the
script to jump to the end by sending `GOODBYE` and making the server shut down
this way.
Therefore, I'm introducing a new line type: auto lines. They look like this

    A: RESET

This line will match the `RESET` message regardless of the fields attached to
the message (just like `!: AUTO` does and send the auto response (the same that
`!: AUTO` sends for each message).

Moreover, I introduce shorthand notation:

    ?: RESET

    *: RESET

    +: RESET

which will internally be expanded to

    {?
        A: RESET
    ?}

    {*
        A: RESET
    *}

    {+
        A: RESET
    +}

respectively.

### DIFF
--- a/boltstub/channel.py
+++ b/boltstub/channel.py
@@ -104,12 +104,15 @@ class Channel:
             self._buffered_msg = self._consume()
         return self._buffered_msg
 
+    def auto_respond(self, msg):
+        self.log("AUTO response:")
+        self.send_struct(self.bolt_protocol.get_auto_response(msg))
+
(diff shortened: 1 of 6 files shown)


bigmontz (approved):
🤖

PR #205, opened 2021-08-10
Auto line arguments

robsdedude:
Contains a bug fix where auto lines are not correctly parsed.

Plus:
Currently, there are no bolt messages with a variable number of fields. Hence,
the fact that auto lines (e.g. `A: HELLO`) accept any number, type, and value
as message fields has no benefit but the drawback of allowing the driver to send
protocol violating messages without having the StubServer complain about it.

Note: it is still possible to accept any type and value (`A: HELLO "*"`) but the
number of expected fields has to be fixed (one in the previous example, or
`A: HELLO "*" "*"` for two, etc.).

### DIFF
--- a/boltstub/parsing.py
+++ b/boltstub/parsing.py
@@ -300,21 +300,9 @@ class ClientLine(Line):
 
 
 class AutoLine(ClientLine):
-    allow_jolt_wildcard = False
-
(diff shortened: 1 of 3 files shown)


bigmontz (approved):
🕹️

PR #210, opened 2021-08-11
Skip test_should_accept_custom_fetch_size_using_driver_configuration in v3

injectives:


robsdedude (approved):
Makes sense, as there is no fetch size in v3.

PR #212, opened 2021-08-11
Migrate tests from Java Driver to Testkit

injectives:
Migrated tests:
- shouldAllowPullAll -> test_should_pull_all_when_fetch_is_minus_one_using_driver_configuration (Bolt bump from v4 to v4.1)
- shouldThrowCommitErrorWhenTransactionCommit -> test_should_error_on_commit_failure_using_tx_commit
- shouldServerWithBoltV4SupportMultiDb -> test_should_check_multi_db_support
- shouldServerWithBoltV3NotSupportMultiDb -> test_should_check_multi_db_support
- shouldBeAbleHandleNOOPsDuringRunCypher -> test_should_accept_noop_during_records_streaming
- shouldThrowCorrectErrorOnCommitFailure -> test_should_error_on_database_shutdown_using_tx_commit
- shouldThrowCorrectErrorOnRunFailure -> test_should_error_on_database_shutdown_using_tx_run


bigmontz (commented):
I agree with the tests, but I don't know if should be in the routing group since it's not testing any routing specific issue. Some are testing the fetchSize usage, others the misc methods like `supportsMultiDB`, and also some errors related to transactions.

injectives:
These tests were migrated from the 'direct connection' tests in java. Since we already have such category in Testkit, there were placed here. In addition, some tests expect to cover different protocol versions and there is already an option for doing so in this category.
In general, we may want to move out all 'no routing' tests from the routing category during refactoring.

bigmontz (approved):
💯 

Agreed to create a ticket to handle organise the tests which are in the routing and no routing modules to more appropriate modules. In the way, these modules will be only responsible for issues related to routing table acquire and routing context.

PR #214, opened 2021-08-11
Fixing few bits for JS driver be complaint with Testkit

bigmontz:
List of adjustments

* Fixes script missmatch in `test_should_be_able_to_use_current_sessions_after_AuthorizationExpired` by make `reader_return_1_failure_return_2_and_3_succeed.script` accept as `DISCARD` or `RESET` at the end of the result consuption with session.close(). The JS driver was already finish the consumption at that point.

### DIFF
--- a/tests/stub/authorization/scripts/v4x0/reader_return_1_failure_return_2_and_3_succeed.script
+++ b/tests/stub/authorization/scripts/v4x0/reader_return_1_failure_return_2_and_3_succeed.script
@@ -18,8 +18,10 @@ A: HELLO {"{}": "*"}
             S: RECORD [1]
                SUCCESS {"type": "r", "has_more": true}
         +}
-        C: DISCARD {"n": -1}
-        S: SUCCESS {}
+        {*
+            C: DISCARD {"n": -1}
+            S: SUCCESS {}
+        *}
     ----
         C: RUN "RETURN 3 as n" {} {"mode": "r"}
            PULL {"n": "*"}
--- a/tests/stub/authorization/scripts/v4x0/reader_return_1_failure_return_2_and_3_succeed.script
+++ b/tests/stub/authorization/scripts/v4x0/reader_return_1_failure_return_2_and_3_succeed.script
@@ -18,10 +18,10 @@ A: HELLO {"{}": "*"}
             S: RECORD [1]
                SUCCESS {"type": "r", "has_more": true}
         +}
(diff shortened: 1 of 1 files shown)


tests/stub/authorization/scripts/v4x0/reader_return_1_failure_return_2_and_3_succeed.script
```diff
@@ -18,8 +18,10 @@ A: HELLO {"{}": "*"}
             S: RECORD [1]
                SUCCESS {"type": "r", "has_more": true}
         +}
-        C: DISCARD {"n": -1}
-        S: SUCCESS {}
+        {*
+            C: DISCARD {"n": -1}
+            S: SUCCESS {}
+        *}
```
robsdedude:
```suggestion
        {?
            C: DISCARD {"n": -1}
            S: SUCCESS {}
        ?}
```

Multiple `DISCARD`s would be a protocol violation.

tests/stub/authorization/scripts/v4x0/reader_return_1_failure_return_2_and_3_succeed.script
```diff
@@ -18,8 +18,10 @@ A: HELLO {"{}": "*"}
             S: RECORD [1]
                SUCCESS {"type": "r", "has_more": true}
         +}
-        C: DISCARD {"n": -1}
-        S: SUCCESS {}
+        {?
+            C: DISCARD {"n": -1}
+            S: SUCCESS {}
+        ?}
```
robsdedude, line 24: Now that I think about it again, shouldn't it be
```suggestion
        {{
            C: DISCARD {"n": -1}
            S: SUCCESS {}
            ----
            A: RESET
        }}
```
Or maybe actually fail?
If I recall this correctly, JS driver is sending a `RESET` instead of a `DISCARD` in some cases.
https://trello.com/c/xHx42X4N/43-discard-or-reset-on-unfinished-result
And one *could* argue that RESET is actually the wrong thing here, as it aborts the query. Not 100% sure.

bigmontz: Actually, the issue only warns in the log, the test was not failing because JS doesn't care if the reset fails, it will not thrown the error, it will just drop the connection from the pool.

PR #219, opened 2021-08-12
Migrate script added in PR #212 to use auto lines

robsdedude:


tests/stub/routing/scripts/v3_no_routing/optional_hello.script
```diff
@@ -1,6 +1,7 @@
 !: BOLT #VERSION#
-!: AUTO RESET
-!: AUTO GOODBYE
 
+?: GOODBYE
```
injectives, line 3: Is this expected here?

tests/stub/routing/scripts/v4x1_no_routing/optional_hello.script
```diff
@@ -1,6 +1,7 @@
 !: BOLT #VERSION#
-!: AUTO RESET
-!: AUTO GOODBYE
 
+?: GOODBYE
```
injectives, line 3: Is this expected here? (same as the other comment)

tests/stub/routing/scripts/v3_no_routing/writer_yielding_multiple_records.script
```diff
@@ -1,13 +1,8 @@
...
-!: AUTO GOODBYE
 
 C: HELLO {"scheme": "basic", "credentials": "c", "principal": "p", "user_agent": "#USER_AGENT#" #ROUTING# #EXTRA_HELLO_PROPS#}
 S: SUCCESS {"server": "#SERVER_AGENT#", "connection_id": "bolt-123456789"}
-# Added this block for GO
-{?
-    C: RESET
-    S: SUCCESS {}
-?}
+*: RESET
```
injectives, line 5: Wouldn't `?: RESET` be sufficient?

tests/stub/routing/scripts/v3_no_routing/writer_yielding_multiple_records.script
```diff
@@ -29,7 +24,5 @@ S: SUCCESS {"server": "#SERVER_AGENT#", "connection_id": "bolt-123456789"}
        RECORD [7]
        SUCCESS {}
 }}
-{?
-    C: RESET
-    S: SUCCESS {}
-?}
+*: RESET
```
injectives, line 27: Wouldn't ?: RESET be sufficient? (duplicate question)

tests/stub/routing/scripts/v3_no_routing/optional_hello.script
robsdedude: JS seems a bit rude ;) and doesn't even say `HELLO` to figure out if the server has multi db support. Arguable it should because unauthenticated connections will be terminated after a while on a real server, if I'm not mistaken. @bigmontz thoughts on this?

tests/stub/routing/scripts/v3_no_routing/writer_yielding_multiple_records.script
robsdedude: It would probably suffice for all official drivers. Who knows about community drivers (also those that might start using TestKit in the future). Also, we have separate tests in the optimization folder that assert that a driver only sends the bare minimum number of resets. In ever other test, I suggest, we shouldn't make assertions on how often a driver wants to reset as long as the resets happen at an appropriate point in the script.

tests/stub/routing/scripts/v3_no_routing/optional_hello.script
bigmontz: Yes, JS Driver does these checks in `No time, brother` mode. It creates the connection, handshakes, checks the version and closes.

injectives (approved):
LGTM

PR #234, opened 2021-09-01
Adapt read timeout tests to Go driver

fbiville:


tests/stub/configuration_hints/scripts/1_second_exceeds.script
```diff
@@ -3,6 +3,10 @@
 
 C: HELLO {"{}": "*"}
 S: SUCCESS {"server": "Neo4j/4.3.1", "hints": {"connection.recv_timeout_seconds": 1}, "connection_id": "bolt-3"}
+{?
+    C: RESET
+    S: SUCCESS {}
+?}
```
bigmontz: You could add the directive `?: RESET` instead of this two lines. It will translated to the same lines but with a more compact  notation.

fbiville: Thanks, will do.

bigmontz (approved):
🍴

tests/stub/configuration_hints/scripts/1_second_exceeds.script
```diff
@@ -3,6 +3,7 @@
 
 C: HELLO {"{}": "*"}
 S: SUCCESS {"server": "Neo4j/4.3.1", "hints": {"connection.recv_timeout_seconds": 1}, "connection_id": "bolt-3"}
+?: RESET
```
robsdedude, line 6: Could actually have been a `*: RESET` instead.

PR #236, opened 2021-09-02
5th round of migration integration tests from Python

robsdedude:


tests/neo4j/sessionrun.py
```diff
@@ -234,6 +234,25 @@ def test_partial_iteration(self):
...
         self._session1 = None
 
+    def test_simple_query(self):
+        def _test():
+            self._driver.close()
+            self._driver = get_driver(self._backend, userAgent="test")
+            self._session1 = self._driver.session("r", fetchSize=2)
+            result = self._session1.run("UNWIND [1, 2, 3, 4] AS x RETURN x")
+            if consume:
+                result.consume()
```
fbiville: are there assertions we can run on the summary in that case?

tests/neo4j/txfuncrun.py
```diff
@@ -24,6 +24,32 @@ def tearDown(self):
...
 
+    def test_simple_query(self):
+        def _test():
+            self._driver.close()
+            self._driver = get_driver(self._backend, userAgent="test")
+            self._session1 = self._driver.session("r", fetchSize=2)
+            tx = self._session1.beginTransaction()
+            result = tx.run("UNWIND [1, 2, 3, 4] AS x RETURN x")
+            if consume:
+                result.consume()
```
fbiville: same question as above

tests/neo4j/txfuncrun.py
```diff
@@ -24,6 +24,32 @@ def tearDown(self):
         self._driver.close()
         super().tearDown()
 
+    def test_simple_query(self):
+        def _test():
+            self._driver.close()
+            self._driver = get_driver(self._backend, userAgent="test")
+            self._session1 = self._driver.session("r", fetchSize=2)
+            tx = self._session1.beginTransaction()
```
fbiville: is this test in the right file? It's testing an explicit tx, while I thought this file was about transaction functions.

fbiville: I think it's been swapped with the test in `txrun.py` :)

fbiville (approved):
LGTM, just 1 minor glitch

tests/neo4j/txfuncrun.py
robsdedude: Oh yes. Well spotted, thanks!

tests/neo4j/sessionrun.py
robsdedude: We could, but there is `tests.stub.summary.test_summary.TestSummary` which is a whole test suite only for making sure that the summary does what it's supposed to.

robsdedude: :thinking: but a little assertion that roughly the expected thing happened might be a good idea.

fbiville (approved):

PR #237, opened 2021-09-03
Add tests for basic, kerberos, custom, and sso auth

robsdedude:


gjmwoods (approved):

tests/stub/iteration/test_iteration_session_run.py
```diff
@@ -74,7 +75,8 @@ def test_discards_on_session_close(self):
         def test():
             uri = "bolt://%s" % self._server.address
             driver = Driver(self._backend, uri,
-                            types.AuthorizationToken(scheme="basic"))
+                            types.AuthorizationToken("basic",  principal="",
```
robsdedude:
```suggestion
                            types.AuthorizationToken("basic", principal="",
```

tests/stub/authorization/scripts/v4x3/scheme_custom.script
```diff
@@ -0,0 +1,13 @@
+!: BOLT 4.3
+
+A: HELLO {"user_agent": "*", "scheme": "wild-scheme", "principal": "I See Something", "credentials": "You Don't See!", "realm": "And it's blue.", "parameters": {"sky?": "no", "my eyes": 0.1, "da be dee da be daa?": true}, "[routing]": null, "[ticket]": ""}
```
gjmwoods: 😂

tests/stub/authorization/scripts/v4x3/scheme_kerberos.script
```diff
@@ -0,0 +1,13 @@
+!: BOLT 4.3
+
+A: HELLO {"user_agent": "*", "scheme": "kerberos", "principal": "", "ticket": "QmFuYW5hIQ==", "[routing]": null, "[credential]": "", "[realm]": ""}
```
gjmwoods: this is where things get fuzzy. I _think_ the Kerberos plugin expects a value of `ticket` for `scheme` (see https://github.com/neo-technology/kerberos-plugin#authenticating-with-kerberos). Java has `kerberos` for scheme and this would mean it would have always been broken for `AuthTokens.kerberos()`

gjmwoods (changes requested):

tests/stub/authorization/scripts/v4x3/scheme_kerberos.script
robsdedude: After digging into the code of the plugin with Florent, we came to the conclusion that the docs might be outdated. We are pretty sure it expects the token to be in the `credentials` field. Regarding the scheme, we had no luck to follow the control flow. Maybe we should contact Olivia who was pretty active in the plugin project to see if she can clarify how it all hangs together.

robsdedude: Olivia confirmed that it belongs into the `credentials` field. Furthermore, she suggested, that the `scheme` actually doesn't matter at all. I suggest we enforce it to be `kerberos` in TestKit regardless.

gjmwoods (approved):

nutkit/protocol/requests.py
```diff
@@ -66,15 +66,28 @@ def __init__(self, uri, authToken, userAgent=None, resolverRegistered=False,
...
+    scheme == "kerberos"
+        - credentials (str)
+    scheme == "bearer"
+        - credentials (str)
+    further schemes should be handled with a multi-purpose auth API
+    (custom auth)
+        - principal (str)
+        - credentials (str)
+        - realm (str)
+        - parameters (dict[str, Any])
```
```suggestion
        - principal (str, optional)
        - credentials (str, optional)
        - realm (str, optional)
        - parameters (dict[str, Any], optional)
```

PR #240, opened 2021-09-08
Update test matrix

robsdedude:
- Use publicly available 4.3 images
 - Update test matrix to be more comprehensive
 - Run suitable integration tests also on cluster configurations

Backport of https://github.com/neo4j-drivers/testkit/pull/241

### DIFF
--- a/main.py
+++ b/main.py
@@ -45,75 +45,84 @@ test_flags = {
 
 
 def initialise_configurations():
-    configurations = []
-    configurations.append(neo4j.Config(
-        name="4.2-cluster",
-        image="neo4j:4.2-enterprise",
-        version="4.2",
-        edition="enterprise",
-        cluster=True,
-        suite="",  # TODO: Define cluster suite
-        scheme="neo4j",
-        download=None,
-        stress_test_duration=90))
+    def generate_config(version, enterprise, cluster, scheme, stress_test):
+        assert (cluster and scheme == "neo4j"
+                or not cluster and scheme in ("neo4j", "bolt"))
+        edition = "enterprise" if enterprise else "community"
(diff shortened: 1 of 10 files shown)


main.py
```diff
@@ -45,75 +45,84 @@
...
+    configurations = [
+        generate_config(version_, enterprise_, cluster_, scheme_, stress_test_)
+        for (version_, enterprise_, cluster_, scheme_, stress_test_) in (
+            # LTS version
+            # 3.5 servers only support routing scheme if configured as cluster
+            ("3.5",    False,       False,    "bolt",   0),
+            ("3.5",    True,        False,    "bolt",   0),
+            ("3.5",    True,        True,     "neo4j", 60),
+            # not officially supported versions
+            ("4.0",    True,        False,    "neo4j",  0),
```
fbiville, line 98: 4.0 has reached its end of life on July 14 (sorry I did not catch this earlier)
Shall we get rid of 4.0 altogether?

tests/neo4j/suites.py
```diff
@@ -35,51 +35,24 @@
 """
 Suite for Neo4j 4.0
```
fbiville, line 36: same as my question above: should we drop this?

fbiville (approved):

main.py
robsdedude: I'd like to get the opinion of the other team members on this. I feel like it would be ok to drop it.

michael-simons: I would be fine with it, I think the differences in Cypher and Indizes are something that mostly effects things like OGM. 
3.5 we need to keep anyway, right?

fbiville: Yep, 3.5 EOL is November 28, 2021

PR #244, opened 2021-09-08
Remove pull pipelining from run tests

injectives:


robsdedude (approved):
More of this goodness! :rocket:

PR #245, opened 2021-09-09
Add optional qid to iteration tests

injectives:


robsdedude:
That's a good point. But seeing this, I think we need to expand the tests in `tests.stub.optimizations.test_optimizations.TestOptimizations.test_uses_implicit_default_arguments` to cover that it's ideal to not send the `qid` when referring to the most recent result stream.

https://trello.com/c/f9xSFTmP

robsdedude (approved):

PR #248, opened 2021-09-10
Simulate SSR-enabled cluster with NotALeader failure

fbiville:
When a cluster enables SSR, it sends a routing table with a single
entry. When a "cluster-retriable error" such as "not a leader"
error occurs, it may happen that the middleware sitting between
the driver and the cluster keeps the failing connection alive.
Because the routing table always points to the same address, the
driver may reuse the same connection that previously failed and
therefore retry and fail again.

This test makes sure that drivers establish a new connection upon
such errors.

Signed-off-by: Florent Biville <florent.biville@neo4j.com>


tests/stub/retry/scripts/clustering/router_retry_once.script
```diff
@@ -0,0 +1,13 @@
...
+!: AUTO GOODBYE
+!: ALLOW RESTART
+
+C: HELLO {"scheme": "basic", "credentials": "c", "principal": "p", "user_agent": "007", "routing": #HELLO_ROUTINGCTX# #EXTRA_HELLO_PROPS#}
+S: SUCCESS {"server": "Neo4j/4.0.0", "connection_id": "bolt-123456789"}
+{*
+    C: ROUTE #ROUTINGCTX# [] null
+    # TODO: double-check if that routing table works with other drivers (may need to
+    # add extra roles)
+    S: SUCCESS { "rt": { "ttl": 1000, "servers": [{"addresses": ["#HOST#:9003"], "role":"WRITE"}]}}
```
robsdedude:
```suggestion
    C: ROUTE #ROUTINGCTX# [] null
    S: SUCCESS { "rt": { "ttl": 1000, "servers": [{"addresses": ["#HOST#:9001"], "role":"ROUTER"}, {"addresses": [], "role":"READER"}, {"addresses": ["#HOST#:9003"], "role":"WRITE"}]}}
```
Is a more realistic RT you'd seem from servers. Even though a server would never be shown as writer but not as reader if I'm not mistaken.

tests/stub/retry/test_retry_clustering.py
```diff
@@ -182,6 +182,40 @@ def twice(tx):
...
+            path=self.script_path("clustering", "router_retry_once.script"),
+            vars=self.get_vars()
+        )
+        vars = {
+            "#ERROR#": "Neo.ClientError.Cluster.NotALeader",
+        }
+        self._writeServer.start(
+            path=self.script_path("retry_with_fail_after_run_server.script"),
+            vars=vars
+        )
```
robsdedude:
```suggestion
        self._writeServer.start(
            path=self.script_path("retry_with_fail_after_run_server.script"),
            vars={"#ERROR#": "Neo.ClientError.Cluster.NotALeader"}
        )
```
just a tad more compact

robsdedude (commented):
Alternatively, one could use an alternative block to have a single server take all three roles. Something along the lines of
```
!: BOLT 4
!: AUTO HELLO
!: AUTO GOODBYE
!: ALLOW CONCURRENT

{*
    C: RESET
    S: SUCCESS {}
*}
{*
    C: ROUTE #ROUTINGCTX# [] null
    S: SUCCESS { "rt": { "ttl": 1000, "servers": [{"addresses": ["#HOST#:9001"], "role":"ROUTER"}, {"addresses": ["#HOST#:9001"], "role":"READER"}, {"addresses": ["#HOST#:9001"], "role":"WRITE"}]}}
    {*
        C: RESET
        S: SUCCESS {}
    *}
*}
C: BEGIN {}
S: SUCCESS {}
{{  
    C: RUN "RETURN 1" {} {}
    S: FAILURE {"code": "#ERROR#", "message": "<whatever>"}
    {?
        C: PULL "*"
        S: IGNORED
    ?}
    {+
        C: RESET
        S: SUCCESS {}
    +}
----
    C: RUN "RETURN 2" {} {}
    S: SUCCESS {"fields": ["1"]}
    C: PULL "*"
    S: RECORD [1]
       SUCCESS {"type": "r"}
    C: COMMIT
    S: SUCCESS {}
    {*
        C: RESET
        S: SUCCESS {}
    *}
}}
{*
    C: ROUTE #ROUTINGCTX# [] null
    S: SUCCESS { "rt": { "ttl": 1000, "servers": [{"addresses": ["#HOST#:9001"], "role":"ROUTER"}, {"addresses": ["#HOST#:9001"], "role":"READER"}, {"addresses": ["#HOST#:9001"], "role":"WRITE"}]}}
    {*
        C: RESET
        S: SUCCESS {}
    *}
*}
```

tests/stub/retry/scripts/clustering/router_retry_once.script
fbiville: That's not what happens with SSR though, so if that's not an issue with other drivers, I'd prefer to keep the RT with a single entry

robsdedude:
I think we should also test that the driver also closes the closes the connection on `Neo.ClientError.General.ForbiddenOnReadOnlyDatabase` as well as check if there are tests in place that check if the driver keeps connections alive on other retryable errors (given the driver's backend reports the feature `OPT_CONNECTION_REUSE = "Optimization:ConnectionReuse"`).

fbiville:
I think we need to rework `test_retry_ForbiddenOnReadOnlyDatabase` and probably remove `test_retry_NotALeader`since it's replaced by `test_close_connection_on_retriable_cluster_error`.

robsdedude:
This issue will have to be resolved in the SSR implementation.

PR #249, opened 2021-09-14
Add tests for omitting qid when possible

robsdedude:


tests/stub/optimizations/scripts/v4x3/all_default.script
```diff
@@ -8,12 +8,21 @@ S: SUCCESS {"server": "Neo4j/4.3.0", "connection_id": "example-connection-id:1"}
     C: BEGIN {}
     S: SUCCESS {}
     C: RUN "*" {} {}
-    S: SUCCESS {"fields": ["1"]}
+    S: SUCCESS {"fields": ["1"], "qid": 1}
```
fbiville, line 11: is this change needed?

tests/stub/optimizations/test_optimizations.py
```diff
@@ -258,13 +258,15 @@ def test():
...
                     res.next()
+                    res.next()
                 tx.commit()
             else:
                 res = session.run("CYPHER")
                 if consume:
                     res.consume()
                 else:
                     res.next()
+                    res.next()
```
fbiville, line 269: what's the reason to add these extra calls?

tests/stub/optimizations/scripts/v4x3/all_default.script
robsdedude: Yes. Not only that, a server would always send  a `qid` back for `'RUN` in an explicit transaction. It's just a coincidence that all official drivers work without it. It's undefined behavior.

tests/stub/optimizations/test_optimizations.py
robsdedude: [These changes](https://github.com/neo4j-drivers/testkit/pull/249/files#diff-85e838161edae13ea4838bd6ccd544501f997cfa3bd2c2825ddb2d81d28bdd9cR15-R25) go along with it. Before it, the `DISCARD` was never reached for drivers with `PULL`-pipelining as they would always send one `PULL` and exhaust the result early this way. Hence I changed the script to yield 2 records.

PR #252, opened 2021-09-20
Add tests for Result.peek() and Result.single()

robsdedude:
Tests are guarded by feature flags `API_RESULT_PEEK = "Feature:API:Result.Peek"`
and `API_RESULT_SINGLE = "Feature:API:Result.Single"` respectively.

### DIFF
--- a/tests/stub/iteration/test_result_peek.py
+++ b/tests/stub/iteration/test_result_peek.py
@@ -1,0 +1,137 @@
+from contextlib import contextmanager
+
+from nutkit.frontend import Driver
+import nutkit.protocol as types
+from tests.shared import (
+    driver_feature,
+    TestkitTestCase,
+    get_driver_name,
+)
+from tests.stub.shared import StubServer
+
+
+class TestResultPeek(TestkitTestCase):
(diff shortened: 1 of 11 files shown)


tests/stub/iteration/test_result_peek.py
```diff
@@ -0,0 +1,137 @@
...
+            session.close()
+        finally:
+            self._server.reset()
+
+    @driver_feature(types.Feature.API_RESULT_PEEK)
+    def test_result_single_with_0_records(self):
+        with self._session("yield_0_records.script") as session:
+            result = session.run("RETURN 1 AS n")
+            with self.assertRaises(types.DriverError) as exc:
+                result.single()
```
fbiville: I'm not following: shouldn't there be a `peek` call there?
it seems `test_result_peek.py` calls `single` and `test_result_single.py` calls `peek` :)
there is a mix of `@driver_feature` as well

fbiville (changes requested):
just a mix-up in tests between peek and single

fbiville (approved):

PR #253, opened 2021-09-21
Raise error on run asap

robsdedude:
Expect the driver to send `RUN` right after `.run(...)` has been called and then wait until it received `SUCCESS` before returning anything to the user.


tests/stub/session_run/scripts/session_yield_error_on_run.script
```diff
@@ -0,0 +1,13 @@
+!: BOLT 4.4
```
AndyHeap-NeoTech: *Being fussy* Just a thought, the file name differs by using yield. Others just have error_on_run. Maybe worth changing.

AndyHeap-NeoTech (approved):

tests/stub/session_run/scripts/session_yield_error_on_run.script
robsdedude: :+1: thanks, fixed.

PR #255, opened 2021-09-23
pytest naming convention

klobuczek:
It looks like the naming was not complying with some conventions as my IDE was not finding the tests before the change.


robsdedude:
I like the idea of renaming the files. And sooner or later I want to migrate TestKit to pytest anyway. But this change changes the id of tests which driver backends rely on to decide if they want to skip a test or not. So for now it's not that straight forward to rename the files.

PR #256, opened 2021-09-23
Adding tests for SSL config options

robsdedude:


certgen/main.go
```diff
@@ -122,28 +122,49 @@ func main() {
...
 
 	// trustedRoot
-	// Trusted by drivers. trustedRoot.pem should be installed on driver Docker image among trusted
+	// Trusted by drivers. trustedRoot.pem should be installed in driver Docker image among trusted
 	// root CAs.
 	trustedRootCert, trustedRootKey, trustedRootDer := generateRoot(anHourAgo, tenYearsFromNow, "trustedRoot")
 	writeKey(path.Join(basePath, "trustedRoot.key"), trustedRootKey)
+	// customRoot + customRoot2
+	// Not trusted by the drivers by default. customRoot.crt should be installed in driver Docker image
+	// in some place. Testkit will only refer to it with it's name (no path). It will be used to test custom CA driver settings.
```
fbiville:
```suggestion
	// in some place. Testkit will only refer to it with its name (no path). It will be used to test custom CA driver settings.
```

certgen/main.go
```diff
@@ -122,28 +122,49 @@ func main() {
...
 	// should be mounted as /etc/ssl/certs/
-	writeCert(path.Join(basePath, "driver", "trustedRoot.crt"), trustedRootDer)
+	writeCert(path.Join(basePath, "driver", "trusted", "trustedRoot.crt"), trustedRootDer)
+	writeCert(path.Join(basePath, "driver", "custom", "customRoot.crt"), customRootDer)
+	writeCert(path.Join(basePath, "driver", "custom", "customRoot2.crt"), customRoot2Der)
 
 	// trustedRoot_server1
-	// Valid dates with hostname set to something that drivers can connect to from driver
-	// Docker container.
+	// Valid dates with hostname set to something that drivers can connect to from driver // Docker container.
```
fbiville:
```suggestion
	// Valid dates with hostname set to something that drivers can connect to from driver Docker container.
```

nutkit/protocol/feature.py
```diff
@@ -14,6 +14,15 @@ class Feature(Enum):
     # This methods asserts that exactly one record in left in the result stream,
     # else it will raise an exception.
     API_RESULT_SINGLE = "Feature:API:Result.Single"
+    # The driver implements explicit configuration options for SSL.
+    #  - enable / disable SSL
+    #  - verify signature against system store / custom cert / not at all
+    API_SSL_CONFIG = "Feature:API:SSLConfig"
+    # The driver understands bolt+s, bolt+ssc, neo4j+s, and neo4j+ssc schemes
+    # and will configure it's ssl options automatically.
```
fbiville:
```suggestion
    # and will configure its ssl options automatically.
```

nutkit/protocol/requests.py
```diff
@@ -62,6 +63,19 @@ def __init__(self, uri, authToken, userAgent=None, resolverRegistered=False,
...
+        # (bool) whether to enable or disable encryption
+        # field missing in message: use driver default (should be False)
+        if encrypted is not None:
+            self.encrypted = encrypted
+        # None: trust system CAs
+        # [] (empty list): trust any certificate
+        # ["path", ...] (list of strings): custom CA certificates to trust
+        # field missing in message: use driver default (should be system CAs)
+        if trustedCertificates is not None:
+            if trustedCertificates == "None":
```
fbiville, line 83: is there a reason why we would allow `"None"` instead of `None`?

tests/tls/test_explicit_options.py
```diff
@@ -0,0 +1,49 @@
...
+class TestExplicitSslOptions(TestkitTestCase):
+    def setUp(self):
+        super().setUp()
+        self._server = None
+        self._driver = get_driver_name()
+
+    def tearDown(self):
+        if self._server:
+            # If test raised an exception this will make sure that the stub
+            # server is killed and it's output is dumped for analysis.
```
fbiville:
```suggestion
            # server is killed and its output is dumped for analysis.
```

nutkit/protocol/requests.py
fbiville: is this because of `test_explicit_config_and_scheme_config`?

tests/tls/test_explicit_options.py
```diff
@@ -0,0 +1,49 @@
...
+            if get_driver_name() in ["javascript"]:
+                self.assertIs("encryption", exc.exception.msg.lower())
+                self.assertIs("trust", exc.exception.msg.lower())
+            else:
+                self.fail("Add expected error type for driver.")
+
+        self._server = TlsServer("trustedRoot_thehost")
+        for scheme in ("neo4j+s", "neo4j+ssc", "bolt+s", "bolt+ssc"):
+            for encrypted in (True, False):
+                for certs in ("None", [], ["customRoot.crt"]):
```
fbiville, line 46:
```suggestion
                for certs in (None, [], ["customRoot.crt"]):
```
wouldn't work?

fbiville: just to make sure I understand why it fails: none of the settings match the configured CA root of the server, right? (because the driver is either using an insecure connection or the wrong root)

tests/tls/test_secure_scheme.py
```diff
@@ -0,0 +1,179 @@
...
+        else:
+            cert = cert_suffix
+        self._server = TlsServer(cert, **kwargs)
+
+    def test_trusted_ca_correct_hostname(self):
+        """
+        Happy path, the server has a valid server certificate signed by a
+        trusted certificate authority.
+        """
+        self.skip_if_missing_driver_features(*self.feature_requirement)
```
fbiville, line 49: why not use an annotation here (and below)?

tests/tls/test_secure_scheme.py
```diff
@@ -0,0 +1,179 @@
...
+        The certificate authority is ok, hostname is ok but the server certificate
+        has expired. Should not connect on expired certificate.
+        """
+        self.skip_if_missing_driver_features(*self.feature_requirement)
+        for driver_config in self.extra_driver_configs:
+            for scheme in self.schemes:
+                with self.subTest(scheme
+                                  + "-" + str(driver_config)):
+                    self._start_server("thehost_expired")
+                    self.assertFalse(self._try_connect(scheme, "thehost",
```
fbiville, line 68: would it be worth asserting the error as well? (I know it's currently not easy since `try_connect` just "passes" on the error... maybe add an optional assertion callback?)

tests/tls/test_secure_scheme.py
```diff
@@ -0,0 +1,179 @@
...
+                                                       driver_config))
+                self._server.reset()
+
+    def test_trusted_ca_wrong_hostname(self):
+        """
+        Verifies that driver rejects connect if hostnames doesn't match
```
fbiville:
```suggestion
        Verifies that driver rejects connections if host names don't match
```

fbiville: same suggestion below

tests/tls/test_secure_scheme.py
```diff
@@ -0,0 +1,179 @@
...
+        {"encrypted": True, "trustedCertificates": ["customRoot2.crt",
+                                                    "customRoot.crt"]},
+    )
+
+    def _start_server(self, cert_suffix, **kwargs):
+        if "Root_" not in cert_suffix:
+            cert = "customRoot_" + cert_suffix
+        else:
+            cert = cert_suffix
+        self._server = TlsServer(cert, **kwargs)
```
fbiville: nit: maybe add another parameter so that you can reuse `_start_server` in `TestSecureScheme`?

tests/tls/test_tls_versions.py
```diff
@@ -23,15 +24,26 @@ def tearDown(self):
...
     def test_1_1(self):
         if self._driver in ["dotnet"]:
             self.skipTest("TLS 1.1 is not supported")
 
         self._server = TlsServer("trustedRoot_thehost", minTls="1", maxTls="1")
-        self.assertTrue(try_connect(self._backend, self._server,
-                                    "neo4j+s", "thehost"))
+        self.assertTrue(self._try_connect())
 
     def test_1_2(self):
```
fbiville, line 50: are there any tests for TLS 1.3?

nutkit/protocol/requests.py
robsdedude: Yes. I need a way to differentiate between "do what the driver understands as default" and pass in languages equivalent of `None`, `null`, `nil`, ... to the driver config, which should make the driver use the system certificates.

tests/tls/test_explicit_options.py
robsdedude: No, it should fail, because the driver should not allow to use the scheme to configure the SSL behavior while also using the config options to configure the SSL behavior. Those should be mutually exclusive.

tests/tls/test_secure_scheme.py
robsdedude: Because inheritance of the class overrides the required driver feature. So I need to be able to reference `self` or the class I'm in.

tests/tls/test_secure_scheme.py
robsdedude:
```suggestion
        Verifies that driver rejects connection if host name doesn't match
```

tests/tls/test_secure_scheme.py
```diff
@@ -0,0 +1,179 @@
...
+                                  + "-" + str(driver_config)):
+                    self._start_server("thehost")
+                    self.assertFalse(self._try_connect(scheme,
+                                                       "thehostbutwrong",
+                                                       driver_config))
+                self._server.reset()
+
+    def test_untrusted_ca_correct_hostname(self):
+        """
+        Verifies that driver rejects connect if hostnames match but CA isn't
```
robsdedude:
```suggestion
        Verifies that driver rejects connection if host name doesn't match
```

tests/tls/test_tls_versions.py
robsdedude: I don't think so. Might be worth adding. At least for those drivers that support it. Python for instance doesn't.

tests/tls/test_secure_scheme.py
robsdedude: I'll put that onto a Trello card and define it as out of scope of this PR. I need to wrap this work up for now. Other tasks gained higher priority. [internal link](https://trello.com/c/nduz6y6h)

tests/tls/test_tls_versions.py
robsdedude: I'll put that onto a Trello card and define it as out of scope of this PR. I need to wrap this work up for now. Other tasks gained higher priority. [internal link](https://trello.com/c/WaJcLWRd)

tests/tls/test_secure_scheme.py
robsdedude: :+1: did that.

tests/tls/shared.py
```diff
@@ -60,6 +67,8 @@ def connected(self):
         raise Exception("Timeout")
 
     def _dump(self):
+        # self._close_pipes()
+        # return
```
injectives: Should this be removed?

tests/tls/test_self_signed_scheme.py
```diff
@@ -0,0 +1,154 @@
...
+                    self._server = TlsServer("trustedRoot_thehost")
+                    self.assertTrue(self._try_connect(scheme, "thehost",
+                                                      driver_config))
+                self._server.reset()
+
+    def test_trusted_ca_expired_server_correct_hostname(self):
+        """
+        A server certificate signed by a trusted CA but the certificate has
+        expired. Go driver happily connects when InsecureSkipVerify is enabled,
+        same for all drivers ?
```
injectives: Should this comment be updated or is it left as a question intentionally?

robsdedude: I just copy-pasted this. Not sure what the current state of the drivers is :thinking:

injectives (approved):

PR #266, opened 2021-10-06
Impersonation

robsdedude:
* The ROUTE message got changed in protocol version 4.4.
   The new tests are guarded by the feature flag `BOLT_4_4 = "Feature:Bolt:4.4"`.
 * Actual impersonation tests are guarded by the feature flag
   `IMPERSONATION = "Feature:Impersonation"`


tests/stub/routing/scripts/v4x4/router_create_adb_with_bookmarks.script
```diff
@@ -0,0 +1,21 @@
+!: BOLT #VERSION#
+
+C: HELLO {"scheme": "basic", "credentials": "c", "principal": "p", "user_agent": "007", "routing": #ROUTINGCTX#}
+S: SUCCESS {"server": "#SERVER_AGENT#", "connection_id": "bolt-123456789"}
+*: RESET
+C: BEGIN {"db": "system"}
+S: SUCCESS {}
+C: RUN "CREATE database foo" {} {}
```
injectives, line 8: Would it make sense to create `adb` here or update the file name?

robsdedude: I think I didn't write the test (maybe I did :shrug:), but what I like about it is that it implicitly assumes that the driver does not care at all about the content of the query. We could change it to something like `some cypher query that creates adb`. Would indeed make a little more sense with the given file name.

injectives (approved):

PR #267, opened 2021-10-07
Fix token expiration exception verification for JS

bigmontz:
Javascript only checks for the error code and the message.


robsdedude (approved):
But why is there no exception you want/can check for?

bigmontz:
> But why is there no exception you want/can check for?

It's a simple JS error with the code and the message. I can send the prototype if it needed.

bigmontz:
Update tests to also check the exception message like go does.

robsdedude (approved):
:chess_pawn:

PR #270, opened 2021-10-12
Add tests for query types in summary

robsdedude:
Drivers should only accept know query types in the result summary
(`"r"`, `"w"`, `"rw"`, `"s"`) or if the key is missing altogether.


bigmontz (approved):
💯

PR #272, opened 2021-10-13
Adding tests to check for the home database resolution

bigmontz:
The home database resolution is not being covered by other tests and it's an important part of the 4.4 release. This PR adds test to exercise this logic in different scenarios.

* [x] Using Session.run in two different sessions in sequence
* [x] Using Session.runTransaction in two different sessions in sequence
* [x] Using Session.run in two different session in parallel
* [x] Using Session.runTransaction in two different in parallel

This PR also fixes a small error in the stubtest


bigmontz (commented):
🦑 

(A comment because I could not auto-approve the PR)

PR #275, opened 2021-10-21
Driver should surface more exceptions on discovery

robsdedude:


tests/stub/routing/test_routing_v4x4.py
```diff
@@ -1954,6 +1955,31 @@ def test_should_fail_with_routing_failure_on_db_not_found_discovery_failure(self
...
+            "router_yielding_invalid_bookmark_failure.script"
+        )
+
+        session = driver.session('r', database=self.adb, bookmarks=["foobar"])
+        with self.assertRaises(types.DriverError) as exc:
+            result = session.run("RETURN 1 as n")
+            result.next()
+        if get_driver_name() in ['java']:
+            self.assertEqual(
+                'org.neo4j.driver.exceptions.FatalDiscoveryException',
```
injectives: This should be: `org.neo4j.driver.exceptions.ServiceUnavailableException`.
This exception will be caused by: ` org.neo4j.driver.exceptions.ClientException` with this message: `Supplied bookmarks 'foobar' does not conform to pattern {database_id}:{tx_id}.`.

robsdedude: Oh really? I copied this over from what happens on trying to connect to a DB that does not exist (`Neo.ClientError.Database.DatabaseNotFound`).

injectives: That code has a special treatment :)
As discussed in another chat, we probably want to give this a special treatment too to avoid retrying (TBC).

injectives: Updated expected error to `org.neo4j.driver.exceptions.ClientException`.

injectives (approved):

PR #277, opened 2021-10-21
Request Routing Table before close the driver

bigmontz:
Asking for the routing table after close the driver could bring strange behaviours.

Javascript discard the driver after it closed, for example.

Depends on: https://github.com/neo4j/neo4j-javascript-driver/pull/794

### DIFF
--- a/tests/stub/routing/test_routing_v4x4.py
+++ b/tests/stub/routing/test_routing_v4x4.py
@@ -657,7 +657,7 @@ class RoutingV4x4(RoutingBase):
         # TODO remove this block once all languages work
         if get_driver_name() in ['go']:
             self.skipTest("requires investigation")
-        if get_driver_name() in ['go', 'java', 'javascript', 'dotnet']:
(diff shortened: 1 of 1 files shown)


tests/stub/routing/test_routing_v4x4.py
```diff
@@ -846,17 +849,19 @@ def test_should_fail_when_writing_on_writer_that_returns_forbidden_on_read_only_
                 )
             failed = True
         session.close()
-        driver.close()
 
         self.assertNotIn(self._writeServer1.address,
                          driver.getRoutingTable(self.adb).writers)
+
```
robsdedude, line 856: The blank line before `driver.close()` is there in some tests but not in others. I think I prefer having it there. You can also decide to not have it, but please be consistent.

robsdedude (approved):

tests/stub/routing/test_routing_v4x4.py
bigmontz: Done

PR #288, opened 2021-11-01
Assert behaviour when AuthorisationExpired happens when fetching routing table

bigmontz:
The connections to the server should being dropped as it is for read and write.


robsdedude:
@bigmontz what's the state of this PR?

robsdedude:
Too stale. Feel free to re-open if still relevant.

PR #290, opened 2021-11-03
Store driver features as enum entries

robsdedude:
This fixes an error in
`tests.neo4j.test_summary.TestSummary.test_protocol_version_information` where
computation of the expected negotiated bolt version relied on having enum
entires available.

### DIFF
--- a/tests/shared.py
+++ b/tests/shared.py
@@ -121,10 +121,7 @@ def get_driver_features(backend):
         raw_features = set(response.features)
         features = set()
         for raw in raw_features:
-            try:
-                features.add(protocol.Feature(raw))
-            except ValueError:  # ignore features we don't know
-                warnings.warn('Unknown feature "%s"' % raw)
+            features.add(protocol.Feature(raw))
         # TODO: remove this once all drivers manage the TLS feature flags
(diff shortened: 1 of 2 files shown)


tests/shared.py
```diff
@@ -118,26 +118,30 @@ def get_driver_features(backend):
...
         if not isinstance(response, protocol.FeatureList):
             raise Exception("Response is not instance of FeatureList")
-        features = set(response.features)
+        raw_features = set(response.features)
+        features = set()
+        for raw in raw_features:
+            try:
+                features.add(protocol.Feature(raw))
+            except ValueError:  # ignore features we don't know
+                warnings.warn('Unknown feature "%s"' % raw)
```
bigmontz: We should fail hard on it. 

It makes easier to stop if we have a typo, for example.

bigmontz (changes requested):
We should fail when some invalid feature name is supplied.

bigmontz (approved):
📚

bigmontz:
Fixing JS driver feature list: https://github.com/neo4j/neo4j-javascript-driver/pull/805

PR #297, opened 2021-11-09
Fix script invalid server messages sequence

bigmontz:
The `routing/v4x1/router_yielding_any_security_failure.script` was cause a invalid scenario by given a `FAILURE` after a `SUCCESS` in response to a `RUN` command.

This sequence could only happen if the driver pipeline the `RUN` followed by `PULL` which is not the case in the script.


robsdedude (approved):
:scream_cat:

bigmontz:
Depends on: https://github.com/neo4j-drivers/testkit/pull/303

PR #301, opened 2021-11-10
Clean up (ab-)using query string for script branching

robsdedude:
I consider it an anti-pattern to use the query string to switch what the stub
server does. This does not apply if
 a) there is no other way of writing the test
 b) you want to avoid starting a new stub-server within a single test.

This pattern should *not* be used to re-use the same script for multiple tests
where it does something different depending on the received query.


bigmontz (approved):
🧹

PR #303, opened 2021-11-10
Decouple test in TestResultList

bigmontz:
Testing two differents scenarios using sub tests are blocking Javascript of using this test suite since it's not compatible with using next and list at the same result.


robsdedude (approved):
:1234:

PR #305, opened 2021-11-15
Add timeout to http integration test

robsdedude:
Depending on the network configuration the connection attempt against the http
endpoint using bolt might not fail immediately but only after the set timeout.
Since that timeout is longer than the TestKit protocol timeout and longer than
we rally should take for one test, we should reduce the timeout in the driver
config.


bigmontz (approved):
⏲️

PR #309, opened 2021-11-15
Capture backend log to the artifacts

bigmontz:
The backend logs were not captured by testkit to the artefacts. This task was being done by the `backend.py` in the driver repository.

The problem with this approach is that sometimes the backend can not have access to the work directory and not be able log it. Also, if there is a bug in the `backend.py` the error logging about what's happening there will not be printed and recorded in the artifacts.


robsdedude (approved):
:spiral_notepad:

PR #310, opened 2021-11-15
Ruby Support

klobuczek:
I have only added special handling where control was not possible with features or test skips.
Still, this resulted in a lot of code duplication but I didn't dare to refactor the code.


tests/shared.py
```diff
@@ -125,9 +125,9 @@ def get_driver_features(backend):
             features.add(protocol.Feature(raw))
         # TODO: remove this once all drivers manage the TLS feature flags
         #       themselves.
-        if get_driver_name() in ["java"]:
+        if get_driver_name() in ["java", "ruby"]:
             features.add(protocol.Feature.TLS_1_1)
-        if get_driver_name() in ["java", "dotnet"]:
+        if get_driver_name() in ["java", "dotnet", "ruby"]:
```
robsdedude: Why do you want it to be added here as opposed to adding the feature flags in your back end? This code was only added to get the PR pipelines green and is supposed to be removed as soon (hopefully sooner than later) as the driver back ends have added it.

nutkit/protocol/responses.py
```diff
@@ -203,7 +203,7 @@ def __eq__(self, _):
                 return True
 
         from tests.shared import get_driver_name
-        if get_driver_name() in ["java", "javascript", "go", "dotnet"]:
+        if get_driver_name() in ["java", "javascript", "go", "dotnet", "ruby"]:
```
robsdedude, line 206: If you driver has the option to provide this info, It'd be the preferred way to send the data. If not, this is fine.

nutkit/protocol/responses.py
```diff
@@ -226,7 +226,7 @@ def __eq__(self, _):
                     "Backend supports well-formatted counter. "
                     "Remove the backwards compatibility check!"
                 )
-        if get_driver_name() in ["java", "javascript", "go", "dotnet"]:
+        if get_driver_name() in ["java", "javascript", "go", "dotnet", "ruby"]:
```
robsdedude: Same here: If you driver has the option to provide this info, It'd be the preferred way to send the data. If not, this is fine.

It should be a rather small change in your back end implementation should the driver already offer the data.

robsdedude (approved):
:gem:

PR #312, opened 2021-11-19
Fix exit code of 0 on failed tests

robsdedude:
When the env had `TEST_RUN_ALL_TESTS=true` and only any combination of
`STUB_TESTS`, `TLS_TESTS`, and `EXTERNAL_TESTKIT_TESTS` was run, TestKit would
not flag any failures with a non-zero exit code.

### DIFF
--- a/main.py
+++ b/main.py
@@ -272,6 +272,11 @@ def main(settings, configurations):
             else:
                 raise
 
+    def _exit():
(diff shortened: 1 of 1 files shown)


klobuczek (approved):

bigmontz (approved):
🖊️

PR #319, opened 2021-11-25
Introduce Support for Websocket in Bolt Stub

bigmontz:
The support for Websockets in the Bolt Stub is need for testing the javascript driver in the browser.

The socket negotiation is done during the `BoltStubRequestHandler.setup` creating the wire using the `create_wire` function. This function handles the WebSocket negotiation and return the Wire with the correct socket negotiated.

The `WebSocket` implements read and receive data using the websockets frames and handle the control frames. This class doesn't not fulfil the whole `rfc6455` implementation because it always treat the payload as binary (which is the only kind of payload we use). The `negotiate_protocol` function also doesn't implement the whole `rfc6455` for negotiate websocket, It doesn't support extensions since it's not needed for BoltStub usecase.

PS: `RegularSocket` and `WebSocket` only implement and test the methods used by `Wire`

### DIFF
--- a/boltstub/tests/test_wiring.py
+++ b/boltstub/tests/test_wiring.py
@@ -184,6 +184,26 @@ class TestWebSocket:
         socket_mock.sendall.assert_called_once()
         socket_mock.sendall.assert_called_with(pong)
 
+    @pytest.mark.parametrize("reserved_control_frame", [
+        (b"\x0B\x00"),
+        (b"\x0C\x00"),
+        (b"\x0D\x00"),
+        (b"\x0E\x00"),
+        (b"\x0F\x00"),
+    ])
+    def test_recv_reserved_control_frame(self, reserved_control_frame, mocker):
+        socket_mock = mocker.Mock()
+        frame = b"\x82\x7E\x00\x7E"
+        payload = randbytes(126)
+        framed_payload = reserved_control_frame + frame + payload
+        socket_mock.recv.side_effect = TestWebSocket.mock_recv(framed_payload)
+
+        websocket = WebSocket(socket_mock)
+
+        received_payload = websocket.recv(1234)
+
+        assert received_payload == payload
+
     def test_recv_multiple_frames(self, mocker):
         payload_list = [
             (randbytes(0), b"\x02\x00"),
--- a/boltstub/tests/test_wiring.py
+++ b/boltstub/tests/test_wiring.py
@@ -6,6 +6,7 @@ import pytest
 
 from ..wiring import (
     create_wire,
+    negotiate_socket,
     RegularSocket,
     WebSocket,
 )
@@ -287,3 +288,16 @@ def test_create_wire(mocker):
 
     assert wire._socket == regular_socket
     wrap_socket.assert_called_with(socket)
+
+
+def test_negotiate_socket_negotiate_regular_socket(mocker):
+    payload = b"\x60\x60\xB0\x17\x00\x00\x01\x04\x00\x00\x00\x04\x00\x00\x00\x03\x00\x00\x00\x00"  # noqa: E501
+    socket = mocker.Mock()
+    socket.recv.return_value = payload
+
+    negotiated_socket = negotiate_socket(socket)
+
+    assert isinstance(negotiated_socket,  RegularSocket)
+    assert negotiated_socket._socket == socket
+    assert negotiated_socket._cache == payload
+    socket.recv.assert_called_once()
--- a/boltstub/tests/test_wiring.py
+++ b/boltstub/tests/test_wiring.py
@@ -160,6 +160,30 @@ class TestWebSocket:
 
         assert received_payload == payload
 
+    @pytest.mark.parametrize("payload,frame", [
+        (randbytes(0), b"\x82\x00"),
+        (randbytes(7), b"\x82\x07"),
+        (randbytes(125), b"\x82\x7D"),
+        (randbytes(126), b"\x82\x7E\x00\x7E"),
+        (randbytes(65535), b"\x82\x7E\xFF\xFF"),
+        (randbytes(65536), b"\x82\x7F\x00\x00\x00\x00\x00\x01\x00\x00"),
+        (randbytes(65537), b"\x82\x7F\x00\x00\x00\x00\x00\x01\x00\x01")
+    ])
+    def test_recv_ping(self, payload, frame, mocker):
+        socket_mock = mocker.Mock()
+        ping = b"\x09\x00"
+        pong = b"\x0A\x00"
+        framed_payload = ping + frame + payload
+        socket_mock.recv.side_effect = TestWebSocket.mock_recv(framed_payload)
+
+        websocket = WebSocket(socket_mock)
+
+        received_payload = websocket.recv(1234)
+
+        assert received_payload == payload
+        socket_mock.sendall.assert_called_once()
+        socket_mock.sendall.assert_called_with(pong)
+
     def test_recv_multiple_frames(self, mocker):
         payload_list = [
             (randbytes(0), b"\x02\x00"),
--- a/boltstub/tests/test_wiring.py
+++ b/boltstub/tests/test_wiring.py
@@ -1,6 +1,6 @@
 
 from functools import reduce
-from random import randbytes
+from random import getrandbits
 
 import pytest
 
@@ -10,6 +10,10 @@ from ..wiring import (
     RegularSocket,
     WebSocket,
 )
+
+
+def randbytes(size):
+    return bytearray(getrandbits(8) for _ in range(size))
 
 
 class TestRegularSocket:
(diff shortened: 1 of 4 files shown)


boltstub/tests/test_wiring.py
```diff
@@ -0,0 +1,338 @@
...
+
+        def test_accessing_the_socket_during_second_call(self, mocker):
+            cache = b"abc"
+            expected = b"barcelos"
+
+            socket_mock = mocker.Mock()
+            socket_mock.recv.return_value = expected
+            regular_socket = RegularSocket(socket_mock, cache)
+
+            _ = regular_socket.recv(1024)
```
robsdedude, line 41:
```suggestion
            regular_socket.recv(1024)
```

robsdedude:
```suggestion
            regular_socket.recv(1024)
```

boltstub/tests/test_wiring.py
```diff
@@ -0,0 +1,338 @@
...
+    wrap_socket.return_value = regular_socket
+
+    wire = create_wire(socket, read_wake_up, wrap_socket)
+
+    assert wire._socket == regular_socket
+    wrap_socket.assert_called_with(socket)
+
+
+def test_negotiate_socket_negotiate_regular_socket(mocker):
+    payload = b"\x60\x60\xB0\x17\x00\x00\x01\x04\x00\x00\x00\x04\x00\x00\x00\x03\x00\x00\x00\x00"  # noqa: E501
```
robsdedude:
```suggestion
    payload = (b"\x60\x60\xB0\x17\x00\x00\x01\x04\x00\x00\x00\x04\x00\x00\x00"
               b"\x03\x00\x00\x00\x00")
```

boltstub/tests/test_wiring.py
```diff
@@ -0,0 +1,338 @@
...
+                    + "Sec-WebSocket-Key: dGhlIHNhbXBsZSBub25jZQ==\r\n"
+                    + "Origin: http://example.com\r\n"
+                    + "Sec-WebSocket-Protocol: chat, superchat\r\n"
+                    + "Sec-WebSocket-Version: 13\r\n\r\n")
+    payload = payload_text.encode("utf-8")
+
+    response_payload_text = ("HTTP/1.1 101 Switching Protocols\r\n"
+                             + "Upgrade: websocket\r\n"
+                             + "Connection: Upgrade\r\n"
+                             + "Sec-WebSocket-Accept: s3pPLMBiTxaQ9kYGzzhZRbK+xOo=\r\n\r\n")  # noqa: E501
```
robsdedude:
```suggestion
    response_payload_text = (
        "HTTP/1.1 101 Switching Protocols\r\n"
        + "Upgrade: websocket\r\n"
        + "Connection: Upgrade\r\n"
        + "Sec-WebSocket-Accept: s3pPLMBiTxaQ9kYGzzhZRbK+xOo=\r\n\r\n"
    )
```

boltstub/wiring.py
```diff
@@ -129,6 +135,101 @@ def __str__(self):
...
+    """
+
+    def __init__(self, _socket) -> None:
+        self._socket = _socket
+
+    def __getattr__(self, item):
+        return getattr(self._socket, item)
+
+    def recv(self, __bufsize) -> bytes:
+        """Receives data from the socket.
```
robsdedude:
```suggestion
        """Receive data from the socket.
```

Use imperative forms for the first line of doc strings.

boltstub/wiring.py
```diff
@@ -129,6 +135,101 @@ def __str__(self):
...
+    This implementation doesn't support extensions
+    """
+
+    def __init__(self, _socket) -> None:
+        self._socket = _socket
+
+    def __getattr__(self, item):
+        return getattr(self._socket, item)
+
+    def recv(self, __bufsize) -> bytes:
```
robsdedude: why does the argument start with `__`?

boltstub/wiring.py
```diff
@@ -129,6 +135,101 @@ def __str__(self):
...
+
+        frame_to_send = bytearray(frame) + bytearray(payload)
+
+        self._socket.sendall(frame_to_send)
+
+        return len(payload)
+
+    def sendall(self, payload):
+        """Send the payload over the socket inside a Websocket frame."""
+        self.send(payload)
```
robsdedude:
```suggestion
    sendall = send
```
for the lazy and idiomatic ones ;)

boltstub/wiring.py
```diff
@@ -251,10 +352,50 @@ def remote_address(self):
         """
         return Address(self.__socket.getpeername())
 
+    @property
+    def _socket(self):
+        return self.__socket
+
```
robsdedude: This looks odd. If a property is marked as internal (i.e. starts with `_`) why even bother exposing it?

boltstub/wiring.py
```diff
@@ -251,10 +352,50 @@ def remote_address(self):
...
 class WireError(OSError):
     """Raised when a connection error occurs."""
 
 
 class BrokenWireError(WireError):
     """Raised when a connection is broken by the network or remote peer."""
+
+
+def negotiate_socket(_socket):
+    def _try_to_negotiate_websocket(__socket, __buffer):
```
robsdedude: About the function name:  
No need to mark the function as internal (starting with `_`). It's scoped to it's parent function so it's not accessible from the outside anyway.

About the parameter names:  
(PEP8)[https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles] is not explicit on this but it states
 * > `_single_leading_underscore`: weak "internal use" indicator. E.g. `from M import *` does not import objects whose names start with an underscore.

 So this only makes sense for module globals as well as class/object attributes/methods.

 * > `single_trailing_underscore_`: used by convention to avoid conflicts with Python keyword, e.g.

So I suggest to use a single trailing underscore to avoid shadowing, e.g., `socket_` instead of `_socket`.


This comment applies to other places as well (e.g. `negotiate_socket(_socket)` just above).

robsdedude (changes requested):
Very valuable change :partying_face:

boltstub/wiring.py
bigmontz: It seems to be like socket define the `__bufsize` argument, I just kept the same name,

boltstub/wiring.py
bigmontz: I exposed it because during the tests I need to check if the Wire was created with the correct socket type.

boltstub/wiring.py
bigmontz: Sure, I will fix up. It was my JS accent talking.

boltstub/wiring.py
```diff
@@ -129,6 +135,99 @@ def __str__(self):
         return "[{}]:{}".format(*self)
 
 
+class RegularSocket:
+    """A socket with an already receive cached value."""
+
+    def __init__(self, socket_, cache_) -> None:
+        self._socket = socket_
+        self._cache = cache_
```
robsdedude:
```suggestion
    def __init__(self, socket_, cache) -> None:
        self._socket = socket_
        self._cache = cache
```
`cache` is not shadowing anything.

boltstub/wiring.py
```diff
@@ -251,10 +350,50 @@ def remote_address(self):
...
+        encoded_key = key.encode(encoding)
+        encrypeted_key = hashlib.sha1(encoded_key).digest()
+        base64_key = base64.standard_b64encode(encrypeted_key).decode(encoding)
+        response = ("HTTP/1.1 101 Switching Protocols\r\n"
+                    + "Upgrade: websocket\r\n"
+                    + "Connection: Upgrade\r\n"
+                    + "Sec-WebSocket-Accept: %s\r\n\r\n") % (base64_key)
+        encoded_response = response.encode(encoding)
+        socket_.sendall(encoded_response)
+        return True
```
robsdedude:
```suggestion
    def try_to_negotiate_websocket(socket__, buffer_):
        encoding = "utf-8"
        encoded = buffer_.strip().decode(encoding)
        headers = encoded.strip().split("\r\n")
        if "Upgrade: websocket" not in headers:
            return False
        for h in headers:
            if "Sec-WebSocket-Key" in h:
                key = h.split(" ")[1]
        key = key + MAGIC_WS_STRING
        encoded_key = key.encode(encoding)
        encrypeted_key = hashlib.sha1(encoded_key).digest()
        base64_key = base64.standard_b64encode(encrypeted_key).decode(encoding)
        response = ("HTTP/1.1 101 Switching Protocols\r\n"
                    + "Upgrade: websocket\r\n"
                    + "Connection: Upgrade\r\n"
                    + "Sec-WebSocket-Accept: %s\r\n\r\n") % (base64_key)
        encoded_response = response.encode(encoding)
        socket__.sendall(encoded_response)
        return True
```

This is an optional suggestion that would avoid shadowing the outer socket_ variable.

robsdedude (approved):
:globe_with_meridians: :electric_plug: :partying_face:

PR #325, opened 2021-12-02
Migrate Java IT test

injectives:
The following IT test has been replaced by stub test:
- `CausalClusteringIT.shouldAllowExistingTransactionToCompleteAfterDifferentConnectionBreaks` -> `Routing.test_should_succeed_when_another_conn_fails_and_discover_using_tx_run`

### DIFF
--- a/tests/stub/routing/test_routing_v4x4.py
+++ b/tests/stub/routing/test_routing_v4x4.py
@@ -2631,3 +2631,67 @@ class RoutingV4x4(RoutingBase):
         self.assertTrue(route_count2 > route_count1 > 0)
         self._writeServer1.done()
         self._readServer1.done()
+
+    def test_should_succeed_when_another_conn_fails_and_discover_using_tx_run(
+            self):
+        # TODO remove this block once fixed
+        if get_driver_name() in ["go"]:
+            self.skipTest("Session close fails with ConnectivityError")
+        # TODO remove this block once fixed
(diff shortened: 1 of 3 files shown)


tests/stub/routing/test_routing_v4x4.py
```diff
@@ -2631,3 +2631,67 @@ def run_and_assert_error(runner):
...
+            self._routingServer1,
+            "router_yielding_writer1_sequentially.script")
+        self.start_server(
+            self._writeServer1,
+            "writer_tx_with_unexpected_interruption_on_run_path.script")
+
+        session1 = driver.session("w", database=self.adb)
+        tx1 = session1.begin_transaction()
+        session2 = driver.session("w", database=self.adb)
+        tx2 = session2.begin_transaction()
```
robsdedude, line 2655: Does the original test also use 2 sessions for this? I'm really curious if 
```python
        session1 = driver.session("w", database=self.adb)
        tx1 = session1.begin_transaction()
        tx2 = session1.begin_transaction()
```

would also work. I'm thinking it should.

robsdedude (approved):
What an intricate test case. LGTM!

tests/stub/routing/test_routing_v4x4.py
injectives: Yes, it used 2 sessions. Java driver prohibits multiple open transactions.

PR #336, opened 2021-12-13
[4.3] Use TeamCity's REST API for downloading artifacts

robsdedude:
This has the main advantage that TestKit does not break every time the nightly
Neo4j builds move to a new patch version.


AndyHeap-NeoTech (commented):
LGTM.

PR #340, opened 2021-12-13
Fixed error in a v4x3 routing script (#335)

AndyHeap-NeoTech:
* Fixed error in a v4x3 routing script

router_and_writer_with_sequential_access_and_bookmark.script had a routing table returned that contained a "db" 
key and value. This is incorrect for the 4.3 protocol, the "db" field was added in 4.4 as part of the impersonation implementation.


robsdedude (approved):
:boot:

PR #341, opened 2021-12-17
[5.0] Add test for in-use stale connection

robsdedude:
The driver should close connections to the same address on encountering
`AuthorizationExpired` on any connection to that address. However, it should
not do this if the connection is still in use by another session/transaction/...

### DIFF
--- a/tests/stub/authorization/test_authorization.py
+++ b/tests/stub/authorization/test_authorization.py
@@ -791,7 +791,7 @@ class TestNoRoutingAuthorization(AuthorizationBase):
     def test_should_drop_connection_after_AuthorizationExpired(self):  # noqa: N802,E501
         self.start_server(
             self._server,
-            "reader_return_1_failure_return_2_and_3_succeed.script"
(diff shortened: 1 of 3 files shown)


tests/stub/authorization/test_authorization.py
```diff
@@ -848,6 +848,53 @@ def test_should_be_able_to_use_current_sessions_after_AuthorizationExpired(  # n
...
+        list(driver.session("r").run("RETURN 3 as n"))
+
+        list(tx1.run("RETURN 4 as n"))
+
+        tx1.commit()
+
+        # now, when session1 has releases its connection, the driver should
+        # remove the connection
+        hangup_count_pre = self._server.count_responses("<HANGUP>")
+        session1.close()
```
bigmontz: You should commit after the count the hangup because the connection could be released just after the commit be sent.

robsdedude: I guess it could :eyes: :mag:

bigmontz (approved):
⛏️

PR #342, opened 2021-12-17
[5.0] Update run_all.py

robsdedude:
* add python async
 * fix java async and reactive (wrong driver name, TestKit only knows `java`)
 * remove .NET mapping 4.4 to 4.3


bigmontz (approved):
🗡️

PR #345, opened 2021-12-17
[4.3] Add test for in-use stale connection

robsdedude:
The driver should close connections to the same address on encountering
`AuthorizationExpired` on any connection to that address. However, it should
not do this if the connection is still in use by another session/transaction/...


bigmontz:
Depends on: https://github.com/neo4j/neo4j-javascript-driver/pull/824

bigmontz (approved):
🍒 ⛏️

PR #347, opened 2021-12-17
[4.4] Fix expected server agent string for aura integration tests

robsdedude:


tests/neo4j/shared.py
```diff
@@ -89,8 +89,10 @@ def __init__(self, version: str, edition: str, cluster: bool):
 
     @property
     def server_agent(self):
-        if self.edition == "aura" and self.version >= "4":
-            return "Neo4j/4.0-aura"
+        if self.edition == "aura":
```
AndyHeap-NeoTech, line 92: Does this mean it will always raise an error against Aura. Perhaps it should skip in this situation?

robsdedude: It will raise an error against aura should a test try to request the assumed agent string. I want to make sure that this one does not slip, so I fail hard... But honestly, it might still slip bcs. aura tests are not part of our PR pipeline...

AndyHeap-NeoTech (approved):

PR #351, opened 2022-01-05
Introduce Support for Websocket in Bolt Stub (#319)

bigmontz:
The support for Websockets in the Bolt Stub is need for testing the javascript driver in the browser.

The socket negotiation is done during the `BoltStubRequestHandler.setup` creating the wire using the `create_wire` function. This function handles the WebSocket negotiation and return the Wire with the correct socket negotiated.

The `WebSocket` implements read and receive data using the websockets frames and handle the control frames. This class doesn't not fulfil the whole `rfc6455` implementation because it always treat the payload as binary (which is the only kind of payload we use). The `negotiate_protocol` function also doesn't implement the whole `rfc6455` for negotiate websocket, It doesn't support extensions since It isn't needed for BoltStub use case.

PS: `RegularSocket` and `WebSocket` only implement and test the methods used by `Wire`

### DIFF
--- a/boltstub/__init__.py
+++ b/boltstub/__init__.py
@@ -36,15 +36,14 @@ import traceback
 from .addressing import Address
 from .channel import Channel
 from .errors import ServerExit
-from .packstream import PackStream
 from .parsing import (
     parse_file,
     Script,
     ScriptFailure,
 )
 from .wiring import (
+    create_wire,
     ReadWakeup,
-    Wire,
 )
 
 log = getLogger(__name__)
@@ -106,7 +105,7 @@ class BoltStubService:
             server_address = None
 
             def setup(self):
-                self.wire = Wire(self.request, read_wake_up=True)
+                self.wire = create_wire(self.request, read_wake_up=True)
                 self.client_address = self.wire.remote_address
                 self.server_address = self.wire.local_address
                 log.info("[#%04X>#%04X]  S: <ACCEPT> %s -> %s",
@@ -114,7 +113,7 @@ class BoltStubService:
                          self.server_address.port_number,
                          self.client_address, self.server_address)
 
-            def handle(self):
+            def handle(self) -> None:
                 with service.actors_lock:
                     actor = BoltActor(deepcopy(script), self.wire)
                     service.actors.append(actor)
--- a/boltstub/tests/test_wiring.py
+++ b/boltstub/tests/test_wiring.py
@@ -1,4 +1,3 @@
-
 from functools import reduce
 from random import getrandbits
 
@@ -47,8 +46,8 @@ class TestRegularSocket:
     class TestWhenCacheIsEmpty:
 
         @pytest.mark.parametrize("empty_cache", [
-            (None),
-            (b"")
+            None,
+            b"",
         ])
         def test_accessing_the_socket(self, empty_cache, mocker):
             expected = b"barcelos"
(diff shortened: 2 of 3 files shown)


bigmontz:
Cherry-picks #319

boltstub/__init__.py
```diff
@@ -106,15 +105,15 @@ class BoltStubRequestHandler(BaseRequestHandler):
...
+                self.wire = create_wire(self.request, read_wake_up=True)
                 self.client_address = self.wire.remote_address
                 self.server_address = self.wire.local_address
                 log.info("[#%04X>#%04X]  S: <ACCEPT> %s -> %s",
                          self.client_address.port_number,
                          self.server_address.port_number,
                          self.client_address, self.server_address)
 
-            def handle(self):
+            def handle(self) -> None:
```
robsdedude, line 116: Type hints :0 :ok_hand:

boltstub/tests/test_wiring.py
```diff
@@ -0,0 +1,341 @@
+
+from functools import reduce
```
robsdedude:
```suggestion
from functools import reduce
```

boltstub/tests/test_wiring.py
```diff
@@ -0,0 +1,341 @@
...
+            actual = regular_socket.recv(1024)
+
+            assert actual == expected
+            socket_mock.recv.assert_called_with(1024)
+
+    class TestWhenCacheIsEmpty:
+
+        @pytest.mark.parametrize("empty_cache", [
+            (None),
+            (b"")
```
robsdedude:
```suggestion
            None,
            b"",
```
No need for parentheses + I'm a fan of trailing commas to reduce git diffs

robsdedude: There are more places like this that I will not comment on.

boltstub/tests/test_wiring.py
```diff
@@ -0,0 +1,341 @@
...
+        wrapped_method = getattr(regular_socket, method_name)
+        original_method = getattr(socket_mock, method_name)
+
+        assert wrapped_method == original_method
+
+
+def duplicate(obj, times):
+    if times == 1:
+        return obj
+    return duplicate(obj + obj, times - 1)
```
robsdedude: FYI: function calls in Python are relatively costly. You'd probably be more efficient with a loop or a reduce.

```python
for _ in range(times):
    obj += obj
```

or 

```python
from functools import reduce
from operator import add

reduce(add, (obj for _ in range (times)))
```

Now, since this is only for testing and times will (probably) not be huge, it won't really make a difference. Just wanted to mention it. Also, Python limits the depth of call stacks by default to something like 1000 or so. So recursion isn't generally my go-to paradigm in Python unless it makes things much easier.

However, I can't find, where this function is used anyway?

bigmontz: This function should be a leftover of previous implementation of some test and I was probably trying to duplicate something 32 or 64 times, or something like this. It was probably for generating long byte arrays.

robsdedude (approved):
:globe_with_meridians: :electric_plug:

boltstub/tests/test_wiring.py
bigmontz: Function calls in JS is almost cost free, most of time it is translated to inline code. The same happens with Java and C++ (in this case, you have to give to the compiler some hints).

PR #353, opened 2022-01-07
Drop server 4.0 from integration test matrix

robsdedude:


main.py
```diff
@@ -98,7 +98,6 @@ def generate_tc_config(version, enterprise, cluster, scheme, stress_test):
             ("3.5",    True,        False,    "bolt",   0),
             ("3.5",    True,        True,     "neo4j", 60),
             # not officially supported versions
-            ("4.0",    True,        False,    "neo4j",  0),
```
fbiville, line 101: this removes only 4.0, not 4.1 (which should be removed as well)

fbiville (changes requested):

main.py
robsdedude: I don't think so. 4.4 drivers were built to work with 4.1 servers and should still be tested against them.

EDIT: Also, I suggested to have a feature freeze on TestKit 4.4. 4.0 ITs shouldn't have been there from the start while 4.1 should have.

fbiville: I'm not sure I'm following the reasoning: why would we remove 4.0 but not 4.1? Both reached [EOL](https://neo4j.com/developer/kb/neo4j-supported-versions/).

fbiville: needs some more internal decision first for 4.1. Let's consider this resolved for now.

fbiville (approved):

PR #354, opened 2022-01-07
Update integration test matrix

robsdedude:
Remove 3.5, 4.0, and 4.1 as they will be EOL when 5.0 goes GA.

Bumped the nightly versions we test against.


fbiville:
Note: 4.0 and 4.1 are already EOL and can be removed on all branches > 4.1, I suppose?
EDIT: just saw the other PRs now :)

fbiville (approved):

PR #365, opened 2022-01-18
Add API_TRANSACTION_CLOSE feature

injectives:


nutkit/protocol/feature.py
```diff
@@ -26,6 +26,8 @@ class Feature(Enum):
     # ...+s: enforce SSL + verify  server's signature with system's trust store
     # ...+ssc: enforce SSL but do not verify the server's signature at all
     API_SSL_SCHEMES = "Feature:API:SSLSchemes"
+    # The driver offers a method to close transaction.
```
robsdedude:
```suggestion
    # The driver offers a method to close transactions. This method should be
    # a no-op if the session is already closed. Else, it should roll back
    # the transaction.
```

robsdedude (approved):
:arrows_clockwise: :heavy_check_mark:

PR #366, opened 2022-01-18
Fix TestMaxConnectionPoolSize side effects in Javascript/Firefox

bigmontz:
The high number of connections created and closed to the same address during theses tests trigger security policies in Firefox, which blocks connection for the address for some minutes.

The avoid this impact in other tests, the port of the server in these test suite was change to 9999 (which is not used in other tests)

### DIFF
--- a/tests/stub/driver_parameters/test_max_connection_pool_size.py
+++ b/tests/stub/driver_parameters/test_max_connection_pool_size.py
@@ -16,7 +16,7 @@ class TestMaxConnectionPoolSize(TestkitTestCase):
 
     def setUp(self):
         super().setUp()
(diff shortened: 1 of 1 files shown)


robsdedude (approved):
:heavy_check_mark: 
Feels like a good enough work-around for now. Not sure future-proof though. There is nothing keeping people from using this port for other tests later.

bigmontz:
> ✔️ Feels like a good enough work-around for now. Not sure future-proof though. There is nothing keeping people from using this port for other tests later.

Ideally, I should resolve it in the backend side, maybe replacing firefox for chromium or disabling the security check in the browser, but I didn't find any configuration property.

PR #372, opened 2022-01-19
Remove some java skips in datatypes, summary, tx fn, routing and version tests

injectives:


tests/stub/routing/test_routing_v4x4.py
```diff
@@ -695,11 +695,11 @@ def test_should_fail_when_writing_on_unexpectedly_interrupting_writer_on_pull_us
         )
 
     def _should_fail_when_writing_on_unexpectedly_interrupting_writer_using_tx_run(  # noqa: E501
```
robsdedude, line 697: I've been tinkering around with this method. What do you think about the following? It's a little more restrictive and still works with all our pipelines (locally tested).

```python
    def _should_fail_when_writing_on_unexpectedly_interrupting_writer_using_tx_run(  # noqa: E501
            self, interrupting_writer_script, fails_on_next=False):
        # TODO remove this block once all languages work
        if get_driver_name() in ["go"]:
            self.skipTest("requires investigation")
        if get_driver_name() in ["go", "dotnet"]:
            self.skipTest("needs routing table API support")
        driver = Driver(self._backend, self._uri_with_context, self._auth,
                        self._userAgent)
        self.start_server(self._routingServer1, "router_adb.script")
        self.start_server(self._writeServer1, interrupting_writer_script)

        session = driver.session("w", database=self.adb)
        tx = session.begin_transaction()
        failed = False
        pipelining_driver = self.driver_supports_features(
            types.Feature.OPT_PULL_PIPELINING
        )
        # TODO: It will be removed as soon as JS Driver
        #       has async iterator api
        if get_driver_name() in ["javascript"] or not pipelining_driver:
            fails_on_next = True
        if fails_on_next:
            result = tx.run("RETURN 1 as n")
                result.next()
        else:
                tx.run("RETURN 1 as n")

        if get_driver_name() in ["java"]:
            self.assertEqual(
                "org.neo4j.driver.exceptions.SessionExpiredException",
                exc.exception.errorType
            )
        session.close()

        self.assertNotIn(self._writeServer1.address,
                         driver.get_routing_table(self.adb).writers)

        driver.close()

        self._routingServer1.done()
        self._writeServer1.done()
```

robsdedude (approved):
:coffee:

PR #378, opened 2022-01-25
Update test_versions for java driver

injectives:


tests/stub/versions/test_versions.py
```diff
@@ -91,9 +91,10 @@ def _run(self, version, server_agent=None, check_version=False,
                         self.assertEqual(summary.server_info.agent,
                                          vars_["#SERVER_AGENT#"])
                     if check_server_address:
-                        self.assertEqual(
-                            summary.server_info.address,
-                            get_dns_resolved_server_address(self._server)
+                        self.assertTrue(
+                            summary.server_info.address in
+                            [get_dns_resolved_server_address(self._server),
+                             self._server.address]
```
robsdedude:
```suggestion
                        self.assertIn(
                            summary.server_info.address,
                            [get_dns_resolved_server_address(self._server),
                             self._server.address]
```
Works too, just fyi. Might make for a better to read failure message, but that should be about all differences.

robsdedude (approved):
:japan:

PR #382, opened 2022-01-26
Improvements for stub server

thelonelyvulpes:
- When stub servers start they will check if the port is in use on the host. (cached as this uses a lot of time)
- When stub servers throw an exception they will include their call stack
- When disconnecting a stubServers socket, use detach on windows
    - https://docs.python.org/3/library/socket.html#socket.socket.detach

### DIFF
--- a/boltstub/__init__.py
+++ b/boltstub/__init__.py
@@ -167,7 +167,9 @@ class BoltStubService:
         try:
             self.server.serve_forever()
         except OSError:
-            print("caught OS Error")
+            msg = ("caught OS Error in stub._server_forever" +
+                   "this may be expected")
+            print(msg)
 
     def _stop_server(self):
         if self.script.context.restarting or self.script.context.concurrent:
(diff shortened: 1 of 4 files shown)


boltstub/__init__.py
```diff
@@ -154,14 +154,21 @@ def finish(self):
...
 
     def _close_socket(self):
         self.server.socket.close()
 
+    def _serve_forever(self):
+        try:
+            self.server.serve_forever()
+        except OSError:
+            msg = "caught OS Error in stub._server_forever"
+            print(msg)
```
robsdedude, line 170: a) you could inline the string :shrug: 
b) more importantly: why do we want to silently ignore OS errors?

tests/stub/shared.py
```diff
@@ -28,6 +29,8 @@
     INTERRUPT_EXIT_CODE = -signal.SIGINT
     POPEN_EXTRA_KWARGS = {}
 
+clean_ports = set()
```
robsdedude, line 32: Would not detect trying to bind two stub servers to the same port. Not sure if needed :shrug:

thelonelyvulpes:
I have a feeling my usage was the cause of concerns so closing. can re-open if needed

PR #384, opened 2022-01-27
Update tests for 5.0 handshake

fbiville:
**⚠️ do not merge yet ⚠️** 
This PR removes the 4.0 Bolt feature flag. If that PR is merged first, drivers will send the 4.0 flag which will be deemed invalid and cause an error.
Therefore, merge the following PRs first:
 - [x] https://github.com/neo4j/neo4j-javascript-driver/pull/852
 - [x] https://github.com/neo4j/neo4j-python-driver/pull/650
 - [x] https://github.com/neo4j/neo4j-java-driver/pull/1135
 - [x] https://github.com/neo4j/neo4j-dotnet-driver/pull/582
 - [x] https://github.com/neo4j/neo4j-go-driver/pull/308


tests/neo4j/shared.py
```diff
@@ -149,13 +149,7 @@ def wrapper(*args, **kwargs):
...
 
 
-def requires_min_bolt_version(feature):
-    if not isinstance(feature, protocol.Feature):
-        raise TypeError("The arguments must be instances of Feature")
-    if not feature.name.startswith("BOLT_"):
-        raise ValueError("Bolt version feature expected")
-
-    min_version = feature.value.split(":")[-1]
+def requires_min_bolt_version(min_version):
```
robsdedude, line 152: I'm not sure about this change. I think having a string instead of a feature flag as argument makes it easier to screw up by accidentally providing an invalid bolt version, doesn't it?

tests/stub/authorization/test_authorization.py
```diff
@@ -519,10 +518,10 @@ def work(tx):
...
             session.read_transaction(work, hooks={
                 "on_send_RetryableNegative": lambda _:
-                    self.switch_unused_servers(
-                        (self._read_server1, self._read_server2),
-                        "reader_tx.script"
-                    )
+                self.switch_unused_servers(
+                    (self._read_server1, self._read_server2),
+                    "reader_tx.script"
+                )
```
robsdedude, line 524: Both are not ideal. The new one I'd argue is even worse.
Usually, you want to avoid multi-line lambdas all together. Consider defining an inner function instead.

Affects other places, too.

tests/stub/authorization/test_authorization.py
```diff
@@ -773,11 +784,12 @@ def tearDown(self):
...
     def get_vars(self):
         return {
-            "#VERSION#": "4.0"
+            "#VERSION#": "4.4"
         }
 
     @driver_feature(types.Feature.OPT_AUTHORIZATION_EXPIRED_TREATMENT)
-    def test_should_drop_connection_after_AuthorizationExpired(self):  # noqa: N802,E501
+    def test_should_drop_connection_after_AuthorizationExpired(
+        self):  # noqa: N802,E501
```
robsdedude: `# noqa: N802,E501` disables two linter checks
 * `E501` check line length (<= 79 characters)
 * `N802` checks that methods use snake case.
 
 I was sorta ok with this method using some Pascal case because that's the name of the error. I wouldn't mind, however, if it was changed to all snake case. Might be better, stylistically speaking.  The line was not too long until I added the `# noqa: N802`. Since this comment cannot go on another line than the one for which it should disable the check, I also had to tell the linter to allow the over-long line.
 
 With the line break the linter will complain about the methods not being snake case.
 
 Affects other places in this PR as well.

tests/stub/routing/test_routing_v5x0.py
```diff
@@ -0,0 +1,2857 @@
...
+    driver_feature,
+    get_dns_resolved_server_address,
+    get_driver_name,
+    get_ip_addresses,
+)
+
+from ._routing import RoutingBase
+
+
+class RoutingV5x0(RoutingBase):
```
robsdedude, line 16: I assume you copied and adopted the tests from `RoutingV4x4`. That's fine but I'd like to see `RoutingV4x4` be a child class of `RoutingV5x0` which changes only the necessary methods and explicitly defines all other methods and make them delegate to `RoutingV5x0`. Have a look at `RoutingV4x3` for an example.

tests/neo4j/shared.py
fbiville: The problem is that I wanted to get rid of the feature flag for Bolt 4.0 so that drivers cannot use it anymore.
Re-introducing the feature flag just for this function seemed a bit overkill.

robsdedude: But why not mark the test as 4.1 minimum? We wouldn't test 4.0 anyway?

fbiville: It would work indeed. The reason I did not initially do it is because I also see this as a self-documenting feature. `requires_min_bolt_version` would give the first version a feature was introduced.
We can of course ditch that aspect completely and just use the next available feature flag.

robsdedude: :thinking: that's a good point. Let's leave it like this and hope it does not get removed once we drop support for all server versions that don't support a given feature.

tests/stub/authorization/test_authorization.py
fbiville: Since this predates the PR, could we tackle this in another one? (not to mention my Python-fu is below the sea level).

tests/stub/routing/test_routing_v5x0.py
fbiville: Do I really need to redeclare all the test methods even though they just call super? That's a bit strange to me.

fbiville: Done anyway :)

tests/stub/iteration/scripts/v4x4/disconnect_on_pull.script
```diff
@@ -1,5 +1,4 @@
-!: BOLT 4
-
```
robsdedude, line 2: Why did you remove the blank line between the bang lines and the script body? It's not necessary but I think it helps with readability (affects many files, will only comment once)?

tests/stub/bookmarks/test_bookmarks_v5.py
```diff
@@ -0,0 +1,154 @@
+from nutkit import protocol as types
+from nutkit.frontend import Driver
+from nutkit.protocol import AuthorizationToken
+from tests.shared import TestkitTestCase
+from tests.stub.shared import StubServer
+
+
+# Tests bookmarks from transaction
+class TestBookmarksV5(TestkitTestCase):
+
```
robsdedude, line 10: In `tests/stub/authorization/test_authorization.py` and maybe other files, this PR changes the code style to not have this blank line at the start of classes. Personally, I also prefer not having it, but both styles are acceptable and I think most places in TestKit have it. Either way, consistency is nice to have.

robsdedude (approved):

tests/stub/bookmarks/test_bookmarks_v5.py
fbiville: The PyCharm gift that keeps on giving 😭

tests/stub/iteration/scripts/v4x4/disconnect_on_pull.script
fbiville: no idea, fixing now

bigmontz (approved):
👍🏼

PR #386, opened 2022-01-31
Add feature flag check for optimized default parameters

thelonelyvulpes:


AndyHeap-NeoTech (approved):
LGTM

robsdedude:
Side note: in fact, I think that the routing tests should be adjusted to not require minimal communication (or at least have two variants, one which enforces implicit defaults - used when the back end reports such capabilities, and, if it doesn't, TestKit should just allow, optionally, all default values).

This can and should be in a separate PR.

PR #388, opened 2022-02-01
javascript: Enable 'Temporary:DriverMaxConnectionPoolSize'

bigmontz:
The `test_should_enforce_pool_size_per_cluster_member` should try to consume the result for getting the error since
javascript driver postpone the error raise.

### DIFF
--- a/tests/stub/routing/test_routing_v4x4.py
+++ b/tests/stub/routing/test_routing_v4x4.py
@@ -2813,7 +2813,8 @@ class RoutingV4x4(RoutingBase):
         session2 = driver.session("w", database=self.adb)
 
         with self.assertRaises(types.DriverError) as exc:
-            session2.begin_transaction()
+            tx = session2.begin_transaction()
+            list(tx.run("RETURN 1 as n"))
 
         if get_driver_name() in ["java"]:
             self.assertEqual(
--- a/tests/stub/routing/test_routing_v4x4.py
+++ b/tests/stub/routing/test_routing_v4x4.py
@@ -2812,9 +2812,13 @@ class RoutingV4x4(RoutingBase):
 
         session2 = driver.session("w", database=self.adb)
 
-        with self.assertRaises(types.DriverError) as exc:
-            tx = session2.begin_transaction()
-            list(tx.run("RETURN 1 as n"))
+        if self.driver_supports_features(types.Feature.OPT_EAGER_TX_BEGIN):
+            with self.assertRaises(types.DriverError) as exc:
+                session2.begin_transaction()
+        else:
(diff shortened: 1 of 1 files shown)


tests/stub/routing/test_routing_v4x4.py
```diff
@@ -2813,7 +2813,8 @@ def test_should_enforce_pool_size_per_cluster_member(self):
         session2 = driver.session("w", database=self.adb)
 
         with self.assertRaises(types.DriverError) as exc:
```
robsdedude: Firstly, I suggest something along the lines of

```python
        if self.driver_supports_features(types.Feature.OPT_EAGER_TX_BEGIN):
                session2.begin_transaction()
        else:
            tx = session2.begin_transaction()
                list(tx.run("RETURN 1 as n"))
```

But there are still open question I have:
 * Am I understanding it correctly, that `session0.begin_transaction()` and `session1.begin_transaction()` allocate connections but don't do anything with them initially?
 * Is the `list(...)` really necessary? Don't we expect drivers to fire of a query (i.e. do real network communication) as soon as we call `tx.run("...")`?

bigmontz: **About the JS behaviour**

`Session.beginTransaction()` returns a Transaction object immediately. It will only throw an exception for errors related to parameters validation and session state (another transaction already open or session closed). Acquiring connection and then send `BEGIN` will be done async and the result will be stored in the transaction state when the promise get resolved. Since the backend returns, the event-loop get back to work and resolves the schedule tasks, including the connection acquire complete which will implies in the driver sending the `BEGIN` message because of the acquire connection was completed. The same will happen again and we will have the begin transaction eventually resolved. 

Externally, all this future work was not consumed, `Transaction.run` don't wait for the acquire and begin be completed, it schedules the `RUN` message to be sent after the previous steps get finished with success and returns the `Result` immediately. 

`Result` contains a promise for record stream observer, this promise could have any kind of error we got previously. So, for getting this error, we should chain ourselves in this promise chain. It could be done by consuming the result. I choose to consume the result by iterate over the records, but I could fetch the next() and it will also works in the same way.

bigmontz: **About the suggestion**

I didn't put any feature flag selection for doing it differently for drivers which has `OPT_EAGER_TX_BEGIN`, because in this test we are verifying the pool size and if the begin/run are lazy or eager doesn't matter much. This way we could get it simple by checking if it throws the expected exception during any point of `begin/run/consume` procedure.

robsdedude (approved):
:horse_racing:

PR #391, opened 2022-02-02
javascript: Do not use float number which could be casted to integers

bigmontz:
Javascript backend will translated this numbers to integers during the json serialization. This way 1.0 will be serialzed as 1.


robsdedude (approved):
:1234: :zap:

PR #395, opened 2022-02-08
Remove dotnet from SSL Scheme feature check workaround

thelonelyvulpes:
Remove dotnet from SSL Scheme feature check workaround


robsdedude (approved):
:hiking_boot:

PR #398, opened 2022-02-09
Improve dumping of stub server output on script mismatch

robsdedude:
This is still not a guarantee that a script mismatch will be dumped on
`StubServer.reset` as there is a race condition going on that cannot be solved
with how communication between the StubServer and TestKit is currently
implemented, but it should increase the odds of a dump significantly.

Signed-off-by: Antonio Barcelos <antonio.barcelos@neo4j.com>


bigmontz (approved):
⏲️ 

(100 ms of waiting looks fine)

PR #399, opened 2022-02-10
Remove language checks in `TestDisconnects.test_disconnect_on_tx_begin`

bigmontz:
It is not needed anymore since `Go` is the only language deviating


robsdedude (approved):
:raised_hands:

PR #401, opened 2022-02-10
Bump connection acquisition timeout in pool size tests

fbiville:
This is needed as the new behavior of the Go driver (following the
soon-to-be-merged context API work) is strictly exercising the
timeout now and fails in CI (i.e. 10ms are a bit too short for the
CI agent it seems).


robsdedude (approved):
:clock1:

PR #404, opened 2022-02-14
Introduce `TestConnectionAcquisitionTimeoutMs`

bigmontz:
This tests should garantee the correct behaviour of this configuration parameter. The connection acquisition timeout must account for the whole acquisition execution time, whether a new connection is created, an idle connection is picked up instead or we need to wait until the full pool depletes.

### DIFF
--- a/nutkit/protocol/feature.py
+++ b/nutkit/protocol/feature.py
@@ -77,7 +77,7 @@ class Feature(Enum):
     # execution time, whether a new connection is created, an idle connection
     # is picked up instead or we need to wait until the full pool depletes.
     CONNECTION_ACQUISITION_TIMEOUT = \
-        "Feature:Configuruation:ConnectionAcquisitionTimeout"
+        "Feature:Configuration:ConnectionAcquisitionTimeout"
 
     # === OPTIMIZATIONS ===
     # On receiving Neo.ClientError.Security.AuthorizationExpired, the driver
--- a/nutkit/protocol/feature.py
+++ b/nutkit/protocol/feature.py
@@ -71,6 +71,13 @@ class Feature(Enum):
     # If this flag is missing, TestKit assumes that attempting to establish
     # such a connection fails.
     TLS_1_3 = "Feature:TLS:1.3"
+    # The driver configuration connection_acquisition_timeout_ms
+    # should be suported.
+    # The connection acquisition timeout must account for the whole acquisition
+    # execution time, whether a new connection is created, an idle connection
+    # is picked up instead or we need to wait until the full pool depletes.
+    CONNECTION_ACQUISITION_TIMEOUT = \
+        "Feature:Configuruation:ConnectionAcquisitionTimeout"
 
     # === OPTIMIZATIONS ===
     # On receiving Neo.ClientError.Security.AuthorizationExpired, the driver
--- a/tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
+++ b/tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
@@ -85,16 +85,16 @@ class TestConnectionAcquisitionTimeoutMs(TestkitTestCase):
 
     def test_should_encompass_the_handshake_time(self):
         """
-        Handshake tooks longer scenario.
+        Handshake takes longer scenario.
 
         This test scenario tests the case where:
 
         1. the connection acquisition timeout is smaller than
             the connection creation timeout
         2. the connection is successfully created and in due time
-        3. the handshake tooks longer than the connection acquisition timeout
+        3. the handshake takes longer than the connection acquisition timeout
 
-        Then the query could not be executed since the connection acquisition
+        Then the query is not executed since the connection acquisition
         timed out.
         """
         self._server.start(
@@ -121,11 +121,11 @@ class TestConnectionAcquisitionTimeoutMs(TestkitTestCase):
 
         1. the connection acquisition timeout is smaller than
             the connection creation timeout
-        2. the connection is successfully tooks longer than the
+        2. the connection takes longer than the
             acquisition timeout to be created
 
-        Then the query could not be executed since the connection acquisition
-        timed out.
+        Then the query is not executed since the connection acquisition
+        times out.
         """
         auth = types.AuthorizationToken("basic", principal="neo4j",
                                         credentials="pass")
@@ -149,11 +149,11 @@ class TestConnectionAcquisitionTimeoutMs(TestkitTestCase):
 
         1. the connection acquisition timeout is bigger than
             the connection creation timeout
-        2. the connection is successfully tooks longer than the
+        2. the connection is successfully takes longer than the
             connection timeout to be created
 
-        Then the query could not be executed since the connection creation
-        timed out.
+        Then the query is not executed since the connection creation
+        times out.
         """
         auth = types.AuthorizationToken("basic", principal="neo4j",
                                         credentials="pass")
@@ -185,8 +185,8 @@ class TestConnectionAcquisitionTimeoutMs(TestkitTestCase):
         3. the connection pool doesn't have connections available in
             suitable time
 
-        Then the begin transaction could not be executed
-        since the connection acquisition timed out.
+        Then the begin transaction is not executed
+        since the connection acquisition times out.
         """
         self._server.start(
             self.script_path("tx_without_commit_or_rollback.script")
--- a/tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
+++ b/tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
@@ -8,6 +8,21 @@ from tests.stub.shared import StubServer
 
 
 class TestConnectionAcquisitionTimeoutMs(TestkitTestCase):
+    """
+    Connection Acquition Timeout Tests.
+
+    The connection acquisition timeout must account for the
+    whole acquisition execution time, whether a new connection is created,
+    an idle connection is picked up instead or we need to wait
+    until the full pool depletes.
+
+    In particular, the connection acquisition timeout (CAT) has precedence
+    over the socket connection timeout (SCT).
+
+    If the SCT is set to 2 hours and CAT to 50ms,
+    the connection acquisition should time out after 50ms,
+    even if the connection is successfully created within the SCT period.
+    """
 
     required_features = (
         types.Feature.BOLT_4_4,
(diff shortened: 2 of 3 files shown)


tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
```diff
@@ -0,0 +1,116 @@
...
+        self._driver = Driver(self._backend, uri, auth,
+                              connection_acquisition_timeout_ms=2000,
+                              connection_timeout_ms=720000)
+
+        self._session = self._driver.session("w")
+
+        with self.assertRaises(types.DriverError):
+            list(self._session.run("RETURN 1 as n"))
+
+    def test_should_have_priotity_over_the_connection_timeout(self):
```
fbiville:
```suggestion
    def test_should_fail_when_acquisition_timeout_is_reached_first(self):
```

tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
```diff
@@ -0,0 +1,116 @@
...
+        self._driver = Driver(self._backend, uri, auth,
+                              connection_acquisition_timeout_ms=2000,
+                              connection_timeout_ms=720000)
+
+        self._session = self._driver.session("w")
+
+        with self.assertRaises(types.DriverError):
+            list(self._session.run("RETURN 1 as n"))
+
+    def test_should_not_suppress_connection_timeout(self):
```
fbiville:
```suggestion
    def test_should_fail_when_connection_timeout_is_reached_first(self):
```

tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
```diff
@@ -0,0 +1,116 @@
...
+        )
+
+        auth = types.AuthorizationToken("basic", principal="neo4j",
+                                        credentials="pass")
+        uri = "bolt://%s" % self._server.address
+        self._driver = Driver(self._backend, uri, auth,
+                              connection_acquisition_timeout_ms=2000,
+                              connection_timeout_ms=720000)
+
+        self._session = self._driver.session("w")
```
fbiville:
```suggestion
        self._session = self._driver.session("r")
```

fbiville: applies to other tests as well

tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
```diff
@@ -0,0 +1,116 @@
...
+                              max_connection_pool_size=1)
+
+        self._sessions = [
+            self._driver.session("w"),
+            self._driver.session("w"),
+        ]
+
+        self._txs = [self._sessions[0].begin_transaction()]
+
+        with self.assertRaises(types.DriverError):
```
fbiville, line 217: just to be sure I'm following, this would fail after the acquisition timeout is reached before the active connection is freed up?

bigmontz: Exactly. In this scenario, I've set up the connection pool size to `1` and blocked only available connection with a transaction.

fbiville:
Maybe add one test for the nominal case where:

1. the connection acquisition timeout is higher than the connection creation timeout
2. the connection is successfully created and in due time
3. the query runs successfully

nutkit/protocol/feature.py
```diff
@@ -71,6 +71,13 @@ class Feature(Enum):
     # If this flag is missing, TestKit assumes that attempting to establish
     # such a connection fails.
     TLS_1_3 = "Feature:TLS:1.3"
+    # The driver configuration connection_acquisition_timeout_ms
+    # should be suported.
+    # The connection acquisition timeout must account for the whole acquisition
+    # execution time, whether a new connection is created, an idle connection
+    # is picked up instead or we need to wait until the full pool depletes.
+    CONNECTION_ACQUISITION_TIMEOUT = \
+        "Feature:Configuruation:ConnectionAcquisitionTimeout"
```
fbiville:
```suggestion
        "Feature:Configuration:ConnectionAcquisitionTimeout"
```

tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
```diff
@@ -0,0 +1,211 @@
...
+                              connection_acquisition_timeout_ms=10000,
+                              connection_timeout_ms=5000)
+
+        self._session = self._driver.session("r")
+
+        list(self._session.run("RETURN 1 as n"))
+
+    def test_should_encompass_the_handshake_time(self):
+        """
+        Handshake tooks longer scenario.
```
fbiville:
```suggestion
        Handshake takes longer scenario.
```

tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
```diff
@@ -0,0 +1,211 @@
...
+
+        This test scenario tests the case where:
+
+        1. the connection acquisition timeout is smaller than
+            the connection creation timeout
+        2. the connection is successfully created and in due time
+        3. the handshake tooks longer than the connection acquisition timeout
```
fbiville:
```suggestion
        3. the handshake takes longer than the connection acquisition timeout
```

tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
```diff
@@ -0,0 +1,211 @@
...
+
+        Then the query could not be executed since the connection acquisition
```
fbiville:
```suggestion
        Then the query is not executed since the connection acquisition
```
(keeping everything in present tense, for consistency)

tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
```diff
@@ -0,0 +1,211 @@
...
+
+    def test_should_fail_when_acquisition_timeout_is_reached_first(self):
+        """
+        Connection creation bigger then acquisition timeout scenario.
+
+        This test scenario tests the case where:
+
+        1. the connection acquisition timeout is smaller than
+            the connection creation timeout
+        2. the connection is successfully tooks longer than the
```
fbiville:
```suggestion
        2. the connection takes longer than the
```

tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
```diff
@@ -0,0 +1,211 @@
...
+            acquisition timeout to be created
+
+        Then the query could not be executed since the connection acquisition
```
fbiville:
```suggestion
        Then the query is not executed since the connection acquisition
```

tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
```diff
@@ -0,0 +1,211 @@
...
+        timed out.
```
fbiville:
```suggestion
        times out.
```

tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
```diff
@@ -0,0 +1,211 @@
...
+
+    def test_should_fail_when_connection_timeout_is_reached_first(self):
+        """
+        Acquisition timeout bigger then connection creation timeout scenario.
+
+        This test scenario tests the case where:
+
+        1. the connection acquisition timeout is bigger than
+            the connection creation timeout
+        2. the connection is successfully tooks longer than the
```
fbiville:
```suggestion
        2. the connection is successfully takes longer than the
```

tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
```diff
@@ -0,0 +1,211 @@
...
+            connection timeout to be created
+
+        Then the query could not be executed since the connection creation
```
fbiville:
```suggestion
        Then the query is not executed since the connection creation
```

tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
```diff
@@ -0,0 +1,211 @@
...
+        timed out.
```
fbiville:
```suggestion
        times out.
```

tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
```diff
@@ -0,0 +1,211 @@
...
+
+        This test scenario tests the case where:
+
+        1. the connection acquisition timeout is higher than
+            the connection creation timeout
+        2. the connection is successfully created and in due time
+        3. the connection pool doesn't have connections available in
+            suitable time
+
+        Then the begin transaction could not be executed
```
fbiville:
```suggestion
        Then the begin transaction is not executed
```

tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
```diff
@@ -0,0 +1,211 @@
...
+        since the connection acquisition timed out.
```
fbiville:
```suggestion
        since the connection acquisition times out.
```

tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
```diff
@@ -0,0 +1,211 @@
...
+    whole acquisition execution time, whether a new connection is created,
+    an idle connection is picked up instead or we need to wait
+    until the full pool depletes.
+
+    In particular, the connection acquisition timeout (CAT) has precedence
+    over the socket connection timeout (SCT).
+
+    If the SCT is set to 2 hours and CAT to 50ms,
+    the connection acquisition should time out after 50ms,
+    even if the connection is successfully created within the SCT period.
```
fbiville, line 24: unfortunately, this is hard to express with a test, but I think it would be worth adding:
```
The CAT must NOT be replaced by the lowest of the two values (itself and SCT).
Indeed, even if SCT is lower than CAT, there could be situations where the pool takes longer to borrow an _idle_ connection than the SCT. Such a scenario should work as long as the overall acquisition happens within the CAT.
This is unfortunately hard to translate into a test.
```

fbiville (approved):
mostly typos, the rest LGTM

PR #405, opened 2022-02-15
Changes for SSL config tests

thelonelyvulpes:


tests/tls/test_secure_scheme.py
```diff
@@ -115,7 +115,7 @@ class TestTrustSystemCertsConfig(TestSecureScheme):
     schemes = "neo4j", "bolt"
     required_features = types.Feature.API_SSL_CONFIG,
     extra_driver_configs = (
-        {"encrypted": True, "trusted_certificates": None},
+        {"encrypted": True, "trusted_certificates": "None"},
```
thelonelyvulpes, line 118: using None and omitting it result in the same behaviour from test kit.
```
        if trustedCertificates is not None:
            if trustedCertificates == "None":
                self.trustedCertificates = None
            else:
                self.trustedCertificates = trustedCertificates
```

robsdedude (approved):
Good catch

PR #408, opened 2022-02-28
element id tests

thelonelyvulpes:
implemented tests for all data structures surrounding element id.
- Node
- Relationship
- UnboundRelationship
- Path


boltstub/simple_jolt/transformers.py
```diff
@@ -428,9 +457,12 @@ def _decode_simple(value, decode_cb):
...
     def _decode_full(cls, value, decode_cb):
-        if not isinstance(value, list) or len(value) != 5:
-            raise JOLTValueError('Expecting list of length 5 after sigil "%s"'
-                                 % cls.sigil)
+        if not isinstance(value, list):
+            raise JOLTValueError('Expecting list after sigil "%s"' % cls.sigil)
+        if len(value) != 5:
+            return JoltRelationTransformer._decode_full_element(value,
+                                                                decode_cb)
+
```
bigmontz: This `if` doesn't let clear why the flow is being deviated, Maybe this could be re-write as something like

```
return decode_relationship_with_elementid(value) 
      if is_relationship_with_element_id(value) 
      else decode_relationship_without_elementid(value)
```

Or something similar, it is also valid for other decode functions.

boltstub/packstream.py
```diff
@@ -264,27 +338,80 @@ def from_jolt_type(cls, jolt: jolt_types.JoltType):
...
+                    node_idx = node_idx + 1
+                    # nodesDict[str(node.id)] = node
+                    uniq_nodes.append(map_node)
+
             # UnboundRelationship structs
-            rels = []
+
+            rel_idxs = {}
+            rel_idx = 1
             for rel in jolt.path[1::2]:
```
thelonelyvulpes, line 331: difference in how we were serializing path ids

boltstub/simple_jolt/transformers.py
thelonelyvulpes: I've added a check if there's an element id at some given index, hopefully that's a bit more clear.

bigmontz: It solves part of the issue. 

The old validation and construction logic should be also moved to it owns method. This _decode_full is doing too much thing and it could be a bit confusing. 

All my focus reading these methods are going the construction and validation of the type which occurs after the flow change.

thelonelyvulpes: I've replaced with  if has field decode v2 else v1.

bigmontz: 🕺🏼

bigmontz (approved):
💯

thelonelyvulpes:
dotnet failing because there's no implementation of CypherPath on testbackend in 5.0, this is fixed in the element id pr.

robsdedude (approved):
Ok now, from my side.

PR #412, opened 2022-03-03
Adjusting the `no writers` tests

bigmontz:
The tests was setting two different readers in the routing table, but it was not starting all the readers.
This was causing random failures in the tests.

Removing the second reader make the test more predictable and avoid possible flackyness.


fbiville (approved):

bigmontz:
The failing test in the Java driver was not related with the changes at all.

PR #413, opened 2022-03-04
Add tests scenarios for VerifyConnectivity and GetServerInfo

bigmontz:
There were two scenarios not covered by tests.
The first one is the case some readers are down and others are running.
The second case is when there are not readers available.


bigmontz:
Depends on: https://github.com/neo4j/neo4j-javascript-driver/pull/888

robsdedude (approved):
:1st_place_medal:

PR #416, opened 2022-03-08
update test_should_enforce_pool_size_per_cluster_member to be more reliable.

thelonelyvulpes:
update test: test_should_enforce_pool_size_per_cluster_member
- increase connection acquisition timeout (fails regularly in dotnet)
- writeServer2.done call when finished


tests/stub/routing/test_routing_v5x0.py
```diff
@@ -2796,7 +2796,7 @@ def test_should_drop_connections_failing_liveness_check(self):
     @driver_feature(types.Feature.TMP_DRIVER_MAX_CONNECTION_POOL_SIZE,
                     types.Feature.API_CONNECTION_ACQUISITION_TIMEOUT)
     def test_should_enforce_pool_size_per_cluster_member(self):
-        acq_timeout_ms = 100
+        acq_timeout_ms = 2000
```
robsdedude, line 2800: Have you played around with the value? The bigger it is, the longer the test takes. So while we want it to be big enough to not be flaky, we also want it to be as small as possible.

thelonelyvulpes: I did experiment, locally more than 1.5Seconds seemed ideal but I wanted to have some headroom for networking or busy build servers. 
Time wise, setting it to 2000ms increased the time the test took by ~50%.

robsdedude (approved):
LGTM

PR #421, opened 2022-03-11
Use API tokens instead of credentials for TeamCity downloads

robsdedude:


robsdedude:
Go keeps failing because of a flaky test that's already tracked internally.

PR #422, opened 2022-03-11
Increase backend timeout for Neo4j integration tests

robsdedude:
Sometimes servers are just very derpy. Especially, when running on the slower
TeamCity build agents. A Cluster that host its leader takes a good while to
catch itself.

This change should help with the flakiness of
`tests.neo4j.test_bookmarks.TestBookmarks.test_can_handle_multiple_bookmarks`
and others.


robsdedude:
This does not seem to fix the issue. Maybe there is something fishy going on with the server as the test seems only to be flaky when run against a 5.0 nightly cluster.

robsdedude:
Replaced by https://github.com/neo4j-drivers/testkit/pull/433

PR #429, opened 2022-03-15
Fix flaky integration tests

robsdedude:
`tests.neo4j.test_tx_func_run.TestTxFuncRun.test_tx_timeout` could fail for
example when a leader change occurred before the error we want to test for.

### DIFF
--- a/tests/neo4j/test_tx_func_run.py
+++ b/tests/neo4j/test_tx_func_run.py
@@ -202,9 +202,16 @@ class TestTxFuncRun(TestkitTestCase):
             nonlocal exc
             with self.assertRaises(types.DriverError) as e:
                 tx.run("MATCH (a:Node) SET a.property = 2").consume()
-            exc = e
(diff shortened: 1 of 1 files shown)


tests/neo4j/test_tx_func_run.py
```diff
@@ -214,9 +221,9 @@ def update2(tx):
             "w", bookmarks=self._session1.last_bookmarks(), database=db
```
fbiville, line 221: maybe for another PR since it's not related to the change, but why use 2 sessions and manually chain bookmarks when we could use 1 session instead?

fbiville (approved):

tests/neo4j/test_tx_func_run.py
robsdedude: I need two sessions because I'm using two managed transactions at the same time to make sure the server is holding the lock on a node while trying to update that node with another managed tx (i.e., another session.)

PR #431, opened 2022-03-15
Improve docs about and handling of env vars

robsdedude:


README.md
```diff
@@ -32,6 +32,13 @@ Environment variables:
     docker will remove the image and all intermediate parent images.
   * `ARTIFACTS_DIR`  
     Name of the directory into which logs and similar debug output is placed.
+  * `TEST_IN_TEAMCITY`
```
fbiville, line 35: for another PR since this is only about docs, what about deriving `TEST_IN_TEAMCITY` from the presence of `TEAMCITY_API_TOKEN`?

README.md
```diff
@@ -51,13 +58,30 @@ Environment variables used to control how tests are executed:
   * `TEST_NEO4J_PASS`  
     Password used to connect to Neo4j server.
     Defaults to 'pass'
+  * `TEST_NEO4J_SCHEME`  
+    Scheme to build the URI when contacting the Neo4j server, default "bolt".
+  * `TEST_NEO4J_HTTP_PORT`  
+    The port on which the Neo4j server is listening for HTTP requests. This is
+    just used to make sure the driver does not try to speak bolt over this port.
+  * `TEST_NEO4J_VERSION`
```
fbiville, line 66: (again for another PR) Go integration tests are configured in a similar way and I'm wondering if it'd make sense to just dynamically retrieve the version instead?

settings.py
```diff
@@ -17,7 +17,7 @@ class ArgumentError(Exception):
 def build(testkit_path):
     """Build. the context based environment variables."""
     in_teamcity = (os.environ.get("TEST_IN_TEAMCITY", "").upper()
-                   in ("TRUE", "1"))
+                   in ("TRUE", "1", "Y", "YES", "ON"))
```
fbiville, line 20: what about "OUI"? just kidding ;-P

fbiville (approved):

settings.py
robsdedude: :baguette_bread: :fr: :rofl: 

Also `"JA"`.

README.md
robsdedude: So `TEST_IN_TEAMCITY` does also alter the output format. The API token is only required for downloading artifacts... They are tightly coupled indeed and I cannot think of a use case where you'd want one but not the other...

README.md
robsdedude: I think I prefer explicitly stating the server version. Otherwise, you have to rely on the driver to tell you what server version you are testing against. I'd rather rely on the driver to test the driver, if you understand.

PR #435, opened 2022-03-18
7th round of migration integration tests from Python 

robsdedude:


tests/neo4j/datatypes/test_temporal_types.py
```diff
@@ -0,0 +1,1024 @@
...
+                    if (isinstance(val, types.CypherDateTime)
+                            and val.utc_offset_s == 0):
+                        self.skipTest(
+                            "timezone library cannot tell the difference "
+                            "between named UTC and 0s offset timezone"
+                        )
+                self._verify_can_echo(val)
+
+    def test_should_echo_all_timezone_ids(self):
+        tz_ids = (
```
fbiville: not sure all of these will work across all languages
but I guess there is only one way to find out :)

fbiville (approved):

tests/neo4j/datatypes/test_temporal_types.py
robsdedude: I'm not even sure if those are all the server supports :worried: It's just all Python supports - the 4 that the server reported it would not support... Maybe I should've checked elsewhere. I think the Java driver or SDN has tests that go over all timezone ids...

nutkit/backend/backend.py
```diff
@@ -136,6 +136,24 @@ def send_and_receive(self, req, timeout=None, hooks=None):
         self.send(req, hooks=hooks)
         return self.receive(timeout, hooks=hooks)
 
+    def _check_system_support(self, type_, meta):
```
fbiville: do you think it's worth introducing a general system check API now? what kind of checks do you foresee?

tests/neo4j/datatypes/_util.py
```diff
@@ -0,0 +1,607 @@
+TZ_IDS = (
```
fbiville, line 1: how are we going to maintain this over time?

nutkit/backend/backend.py
robsdedude: Maybe ask JS if it knows the difference between `0.0` and `0` 😐. I was just thinking that I don't want to introduce a message just for checking if a certain timezone is supported. That felt way to specific.

tests/neo4j/datatypes/_util.py
robsdedude: I don't quite know. Any suggestions welcome. I mean this list should probably only ever get extended and never reduced.

nutkit/backend/backend.py
fbiville: I usually try to generalize something once I've got evidence of several specializations but YMMV

you could argue we can start with the simple, very specific timezone check and generalize it afterwards
that being said, such a generalization, if done at a later stage, would likely be a breaking change that would affect all testkit clients

robsdedude: Yes. If it was just TestKit that's effected I wouldn't have done this. But as you say it affects the protocol and hence all backends. But I don't have a strong opinion on that. If you have, I can turn this into a specific message pair.

tests/neo4j/datatypes/test_temporal_types.py
robsdedude: Extension happened :cold_sweat:

PR #440, opened 2022-03-28
Define Result#list error mapping for Go

fbiville:


tests/stub/iteration/scripts/v4x4/tx_pull_list.script
```diff
@@ -6,7 +6,7 @@ A: HELLO {"{}": "*"}
 C: BEGIN {"{}": "*"}
 S: SUCCESS {}
 C: RUN "*" "*" "*"
-S: SUCCESS {"fields": ["n"]}
+S: SUCCESS {"fields": ["n"], "qid": 0}
```
fbiville: I guess all other drivers fail when `qid` is there, while `Go` fails without it 🤔

fbiville: the thing I don't get is that `qid` should be present since we're running `BEGIN`+`RUN`: https://7687.org/bolt/bolt-protocol-message-specification-4.html#server-response-success-11

fbiville: Other drivers fail at subsequent `PULL` message, since they don't implement `OPT_IMPLICIT_DEFAULT_ARGUMENTS`

fbiville:
I cannot really decipher the Java failure here...

robsdedude:
> I cannot really decipher the Java failure here...

It fails some TLS test. Most certainly not related to these changes.

tests/stub/iteration/scripts/v4x4/tx_pull_list.script
fbiville: this was a Go driver issue, these additions here are not needed

robsdedude (approved):
:clap:

PR #443, opened 2022-04-01
Remove managed tx lifetime tests

fbiville:
The existing managed transaction APIs (see list below) are just to
be deprecated and behave as in 4.x.

The APIs in question are: Commit/Rollback/Close.

The new managed transaction APIs will not include any operation
like these but that cannot be really tested by TestKit.


AndyHeap-NeoTech (approved):
Perfectly deleted test. No comments to add :)

PR #445, opened 2022-04-05
Introduce Testkit error types

injectives:
The goal of this update is to introduce Testkit error types that Testkit would expect driver backend to map its error to. This would allow Testkit to remain driver agnostic and still verify expected errors.

### DIFF
--- a/nutkit/protocol/error_type.py
+++ b/nutkit/protocol/error_type.py
@@ -7,7 +7,7 @@ class ErrorType(Enum):
     # - interrupting nodes
     # - no writers available
     # - connectivity issues
-    SESSION_EXPIRED_ERROR = "session_expired_error"
+    SESSION_EXPIRED_ERROR = "SessionExpiredError"
 
     # discovery
     # - procedure not found on discovery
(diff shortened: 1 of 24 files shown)


tests/neo4j/test_direct_driver.py
```diff
@@ -75,6 +76,9 @@ def test_fail_nicely_when_using_http_port(self):
         if get_driver_name() in ["python"]:
             self.assertEqual(e.exception.errorType,
                              "<class 'neo4j.exceptions.ServiceUnavailable'>")
+        elif get_driver_name() in ["java"]:
+            self.assertEqual(ErrorType.SERVICE_UNAVAILABLE_ERROR.value,
+                             e.exception.errorType)
```
bigmontz: We could use a feature flag instead of a check for the language name.

injectives: Yes, we can. But we wanted to stop having temporary flags. Once all referenced languages implement these types, we should get rid of these branches altogether.

injectives: I will leave this open for now. Since this is a draft, we can update it later on if we decide to do it.

robsdedude: We could possibly have a permanent feature flag `API:UnifiedErrorClasses` or whatnot. Not 100% sold on that option though.

nutkit/protocol/error_type.py
```diff
@@ -97,40 +98,37 @@ class ErrorType(Enum):
...
-    PROTOCOL_ERROR = "protocol_error"
+    PROTOCOL_ERROR = "ProtocolError"
 
     # TO BE DISCUSSED
     # python throws when transaction has been closed
     # java throws client exception
-    TRANSACTION_ERROR = "transaction_error"
-
-    # TO BE DISCUSSED
-    MANAGED_TRANSACTION_ERROR = "managed_transaction_error"
```
robsdedude: I decided to ditch `MANAGED_TRANSACTION_ERROR = "managed_transaction_error"` as this was only used to signify to the user that they can't manage (e.g., `.commit`) a tx inside a tx function. Since we decided to re-design this and won't/can't have tests for the new approach, there is no sense in having this error class any longer.

nutkit/protocol/error_type.py
```diff
@@ -0,0 +1,134 @@
...
+    # - illegal argument value
+    ILLEGAL_ARGUMENT_ERROR = "IllegalArgumentError"
+
+    # protocol violations
+    PROTOCOL_ERROR = "ProtocolError"
+
+    # TO BE DISCUSSED
+    # python throws when transaction has been closed
+    # java throws client exception
+    TRANSACTION_ERROR = "TransactionError"
```
robsdedude, line 131: This could also very well be the same as `"IllegalStateError"`, I think. It feels very much like it.

nutkit/protocol/error_type.py
injectives: Yes, this makes sense 👍

tests/neo4j/test_tx_run.py
```diff
@@ -324,9 +325,15 @@ def test_tx_timeout(self):
...
-                             "<class 'neo4j.exceptions.TransientError'>")
+        if get_driver_name() in ["java", "python"]:
+            self.assertEqual(ErrorType.TRANSIENT_ERROR.value,
+                             e.exception.errorType)
+        # "Neo.TransientError.Transaction.LockClientStopped"
+        # will be moved to client-error since it is not retriable.
+        # Shall we categorized it as a client error?
+        if get_driver_name() in ["javascript"]:
+            self.assertEqual(ErrorType.CLIENT_ERROR.value,
+                             e.exception.errorType)
```
bigmontz, line 336: /\

PR #374, opened 2022-01-20
Add support for build args for Driver Docker Image

bigmontz:
This feature solves problems like select the version of languages runtimes or libraries present in the driver docker image during the test.
Sending `--build-arg`s to the image makes easier to handle this kind of configuration on the image level instead of changing the version using other scripts in a post process.

The first use-case of this feature will be the selection of Nodejs version in the Javascript Driver.

More info about how to use it in the updated docs.

PR #450, opened 2022-04-19
Add support for build args for Driver Docker Image (#374)

robsdedude:
This feature solves problems like select the version of languages runtimes or libraries present in the driver docker image during the test.
Sending --build-args to the image makes easier to handle this kind of configuration on the image level instead of changing the version using other scripts in a post process.

The first use-case of this feature will be the selection of Nodejs version in the Javascript Driver.

More info about how to use it in the updated docs.

Backport of https://github.com/neo4j-drivers/testkit/pull/374

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,3 +232,23 @@ Environment variables:
     Username used to connect, defaults to neo4j
   * `TEST_NEO4J_PASS`  
     Password used to connect
+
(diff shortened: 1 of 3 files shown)


bigmontz (approved):
🧛🏼

PR #452, opened 2022-04-20
Remove ignored messages from routing tests

thelonelyvulpes:
Remove ignored being sent by server after sending error message in some routing stub scripts.


robsdedude (approved):
:wastebasket:

PR #457, opened 2022-05-06
Fix `test_no_reset_on_clean_connection`

bigmontz:
The test was not exausting the record consumption, so it was not always true the driver processed the SUCCESS when the Session.close is called. Closing a session with a result open will trigger a reset in the javascript driver, which makes the test fail.

### DIFF
--- a/tests/stub/optimizations/test_optimizations.py
+++ b/tests/stub/optimizations/test_optimizations.py
@@ -89,7 +89,7 @@ class TestOptimizations(TestkitTestCase):
                     res.consume()
                 else:
                     results.append(
-                        list(map(lambda r: r.value, res.next().values))
+                        list(map(lambda r: r.values[0].value, res))
                     )
                 if use_tx == "commit":
                     tx.commit()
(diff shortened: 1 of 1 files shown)


tests/stub/optimizations/test_optimizations.py
```diff
@@ -89,7 +89,7 @@ def double_read(self, mode, new_session, use_tx, routing,
                     res.consume()
                 else:
                     results.append(
-                        list(map(lambda r: r.value, res.next().values))
+                        list(map(lambda r: r.values[0].value, res))
```
robsdedude:
```suggestion
                        [value.value for rec in res for value in rec.values]
```

Probably not exactly what the test is about, but I still would want to see if the driver for what ever reason decides to send more or less than one value per result.

tests/stub/optimizations/test_optimizations.py
```diff
@@ -101,7 +101,7 @@ def double_read(self, mode, new_session, use_tx, routing,
...
```
robsdedude:
```suggestion
                        [value.value for rec in res for value in rec.values]
```

robsdedude (dismissed):

PR #460, opened 2022-05-10
Add error checks for 4.4/5.0 changing code

thelonelyvulpes:
Implement error code mapping for error codes that changed.
- what tests could we add?
- could we include if an error is retriable.

### DIFF
--- a/tests/stub/errors/scripts/test_error.script
+++ b/tests/stub/errors/scripts/test_error.script
@@ -1,4 +1,4 @@
-!: BOLT 4.4
+!: BOLT #BOLT_PROTOCOL#
 
 A: HELLO {"{}": "*"}
 *: RESET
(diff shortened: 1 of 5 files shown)


tests/stub/errors/scripts/test_error.script
```diff
@@ -0,0 +1,13 @@
+!: BOLT 4.4
+
+A: HELLO {"{}": "*"}
+*: RESET
+C: RUN { "U": "*"} {"{}": "*"} {"{}": "*"}
+S: FAILURE {"code": "#ERROR_CODE#", "message": "Something went wrong"}
```
thelonelyvulpes, line 6: I'm just doing this on run, but should it really be done in the block of a transaction?

thelonelyvulpes:
Better alternative https://github.com/neo4j-drivers/testkit/pull/462

PR #462, opened 2022-05-12
Verify mapped transient failures are not retried

injectives:
In 5.0 the following failures have been mapped to different category:
- `Neo.TransientError.Transaction.Terminated` -> `Neo.ClientError.Transaction.Terminated`
- `Neo.TransientError.Transaction.LockClientStopped` -> `Neo.ClientError.Transaction.LockClientStopped`

The added tests verify that in event of receiving previous errors (may happen when connecting to pre 5.0 server) the error code gets mapped to the new category and no retries happen.


bigmontz (approved):
🐑 

(The JS failures could be solved by consuming all the records in the tx function. However, it is not needed since the javascript driver is not behaving well anyway)

PR #471, opened 2022-06-10
Fix integration tests to also work with the latest 5.0 drop

robsdedude:
There's lots to discuss about this, I'm afraid.

First the reason why we do this: the Aura drops run integration tests with out drivers. Since 5.0 drops are in the making, but the 5.0 drivers are not yet ready and users that connect to those drops might still be running on an older driver, those ITs run against 4.4 drivers.

Technically, we don't support older drivers connecting to newer servers. But no connection option for the new drops isn't viable either. So here goes: making TestKit 4.4 more compatible with the current dev state of the 5.0 server.


AndyHeap-NeoTech (approved):
Looks good.

PR #475, opened 2022-06-14
Remove Python and Ruby from special LockClientStopped handling

robsdedude:
* Python has a fix prepared:
   https://github.com/neo4j/neo4j-python-driver/pull/742
 * Ruby is not part of our pipelines and hence needs no exception just to get
   our TestKit pipelines green. The maintainers should, however, notice through
   failing tests that there is work required, once they upgrade to 5.0


bigmontz (approved):
🧇

fbiville (approved):

PR #484, opened 2022-06-28
Wait for database management procedures

robsdedude:
This became especially necessary for 5.0 as it changed the async behavior of those procedures.

This will also need backporting to 4.4

### DIFF
--- a/tests/neo4j/shared.py
+++ b/tests/neo4j/shared.py
@@ -178,3 +178,32 @@ def requires_min_bolt_version(min_version):
             return func(*args, **kwargs)
         return wrapper
     return bolt_version_decorator
+
+
+class QueryBuilder:
+    @staticmethod
+    def escape_identifier(identifier):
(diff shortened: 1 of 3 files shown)


tests/neo4j/shared.py
```diff
@@ -178,3 +178,32 @@ def wrapper(*args, **kwargs):
...
+        version = get_server_info().version
+        return "CREATE DATABASE {}{}".format(
+            QueryBuilder.escape_identifier(database),
+            QueryBuilder._wait_clause(version) if wait else ""
+        )
+
+    @staticmethod
+    def drop_db(database, if_exists=True, wait=True):
+        version = get_server_info().version
+        if_exists = " IF EXISTS" if if_exists else ""
```
bigmontz: the change from the `if_exists` boolean to `if_exists` string is a bit awkward and it is making the code a bit more difficult realize.

bigmontz (approved):
🏁

PR #488, opened 2022-06-30
Make sure unknown timezones don't break the driver

robsdedude:
Should be ported back to TestKit 4.4 when merged

### DIFF
--- a/tests/stub/types/scripts/v5x0/echo_unknown_then_known_zoned_date_time.script
+++ b/tests/stub/types/scripts/v5x0/echo_unknown_then_known_zoned_date_time.script
@@ -1,0 +1,37 @@
+!: BOLT 5.0
+
+A: HELLO {"{}": "*"}
+*: RESET
+A: BEGIN {"{}": "*"}
+C: RUN "MATCH (n:State) RETURN n.founded AS dt" {} {"{}": "*"}
+S: SUCCESS {"fields": ["dt"], "qid": 1}
+{{
+    C: PULL {"n": {"Z": "*"}, "[qid]": -1}
+----
+    C: PULL {"n": {"Z": "*"}, "qid": 1}
+}}
+
+# Noe4j land: unknown land with unknown timezone
+# S: RECORD [{"T": "1970-01-01T01:00:00+00:00[Europe/Noe4j]"}]
+# chunk start, 20 bytes
+S: <RAW> 00 14
+# RECORD message
+S: <RAW> B1 71
+# list of fields header (1 long)
+S: <RAW> 91
+# only list entry is the datetime field which is a struct with 3 fields and tag x69
+S: <RAW> B3 69
+# 0 seconds, 0 nanoseconds
+S: <RAW> 00 00
+# tz id string "Europe/Noe4j" (12 chars long, 8C header)
+S: <RAW> 8C 45 75 72 6f 70 65 2f 4e 65 6f 34 6a
+# 0 chunk to end RECORD message
+S: <RAW> 00 00
+
+# Now something easy: a fixed offset timezone
+S: RECORD [{"T": "1970-01-01T01:00:00+00:30"}]
+
+S: SUCCESS {"type": "r"}
+A: COMMIT
+*: RESET
+?: GOODBYE
--- a/tests/stub/types/scripts/v5x0/echo_unknown_then_known_zoned_date_time.script
+++ b/tests/stub/types/scripts/v5x0/echo_unknown_then_known_zoned_date_time.script
@@ -11,8 +11,8 @@ S: SUCCESS {"fields": ["dt"], "qid": 1}
     C: PULL {"n": {"Z": "*"}, "qid": 1}
 }}
 
-# Noe4j land: unknown land with unknown timezone
-# S: RECORD [{"T": "1970-01-01T01:00:00+00:00[Europe/Noe4j]"}]
+# Neo4j land: unknown land with unknown timezone
+# S: RECORD [{"T": "1970-01-01T01:00:00+00:00[Europe/Neo4j]"}]
 # chunk start, 20 bytes
 S: <RAW> 00 14
 # RECORD message
@@ -23,7 +23,7 @@ S: <RAW> 91
 S: <RAW> B3 69
 # 0 seconds, 0 nanoseconds
 S: <RAW> 00 00
-# tz id string "Europe/Noe4j" (12 chars long, 8C header)
+# tz id string "Europe/Neo4j" (12 chars long, 8C header)
 S: <RAW> 8C 45 75 72 6f 70 65 2f 4e 65 6f 34 6a
 # 0 chunk to end RECORD message
 S: <RAW> 00 00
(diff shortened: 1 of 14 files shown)


tests/stub/types/scripts/v5x0/echo_unknown_then_known_zoned_date_time.script
```diff
@@ -0,0 +1,37 @@
...
+A: BEGIN {"{}": "*"}
+C: RUN "MATCH (n:State) RETURN n.founded AS dt" {} {"{}": "*"}
+S: SUCCESS {"fields": ["dt"], "qid": 1}
+{{
+    C: PULL {"n": {"Z": "*"}, "[qid]": -1}
+----
+    C: PULL {"n": {"Z": "*"}, "qid": 1}
+}}
+
+# Noe4j land: unknown land with unknown timezone
```
fbiville:
```suggestion
# Neo4j land: unknown land with unknown timezone
```

tests/stub/types/scripts/v5x0/echo_unknown_then_known_zoned_date_time.script
```diff
@@ -0,0 +1,37 @@
...
+# S: RECORD [{"T": "1970-01-01T01:00:00+00:00[Europe/Noe4j]"}]
```
fbiville:
```suggestion
# S: RECORD [{"T": "1970-01-01T01:00:00+00:00[Europe/Neo4j]"}]
```

tests/stub/types/scripts/v5x0/echo_unknown_then_known_zoned_date_time.script
```diff
@@ -0,0 +1,37 @@
...
+S: <RAW> 00 14
+# RECORD message
+S: <RAW> B1 71
+# list of fields header (1 long)
+S: <RAW> 91
+# only list entry is the datetime field which is a struct with 3 fields and tag x69
+S: <RAW> B3 69
+# 0 seconds, 0 nanoseconds
+S: <RAW> 00 00
+# tz id string "Europe/Noe4j" (12 chars long, 8C header)
```
fbiville:
```suggestion
# tz id string "Europe/Neo4j" (12 chars long, 8C header)
```

tests/stub/types/scripts/v5x0/echo_unknown_zoned_date_time.script
```diff
@@ -0,0 +1,34 @@
...
+A: BEGIN {"{}": "*"}
+C: RUN "MATCH (n:State) RETURN n.founded AS dt" {} {"{}": "*"}
+S: SUCCESS {"fields": ["dt"], "qid": 1}
+{{
+    C: PULL {"n": {"Z": "*"}, "[qid]": -1}
+----
+    C: PULL {"n": {"Z": "*"}, "qid": 1}
+}}
+
+# Noe4j land: unknown land with unknown timezone
```
fbiville:
```suggestion
# Neo4j land: unknown land with unknown timezone
```

tests/stub/types/scripts/v5x0/echo_unknown_zoned_date_time.script
```diff
@@ -0,0 +1,34 @@
...
+# S: RECORD [{"T": "1970-01-01T01:00:00+00:00[Europe/Noe4j]"}]
```
fbiville:
```suggestion
# S: RECORD [{"T": "1970-01-01T01:00:00+00:00[Europe/Neo4j]"}]
```

tests/stub/types/scripts/v5x0/echo_unknown_zoned_date_time.script
```diff
@@ -0,0 +1,34 @@
...
+S: <RAW> 00 14
+# RECORD message
+S: <RAW> B1 71
+# list of fields header (1 long)
+S: <RAW> 91
+# only list entry is the datetime field which is a struct with 3 fields and tag x69
+S: <RAW> B3 69
+# 0 seconds, 0 nanoseconds
+S: <RAW> 00 00
+# tz id string "Europe/Noe4j" (12 chars long, 8C header)
```
fbiville:
```suggestion
# tz id string "Europe/Neo4j" (12 chars long, 8C header)
```

thelonelyvulpes (dismissed):

robsdedude:
Unknown time zones apparently crash the JS backend => skip/fix required.

tests/stub/types/test_temporal_types.py
```diff
@@ -35,54 +35,82 @@ def _create_direct_driver(self):
...
+        self.assertEqual(tx_count, 1)
+
+    def _test_unknown_then_known_zoned_date_time(self, patched):
+        def work(tx):
+            res = tx.run("MATCH (n:State) RETURN n.founded AS dt")
+            with self.assertRaises(types.DriverError) as exc:
+                res.next()
+            self.assertIn("Europe/Neo4j", exc.exception.msg)
+
+            rec = res.next()
```
injectives, line 99: Should we not expect result to be in invalid state by this point given the error above?

robsdedude: I figured it would be good to be as lenient as possible. Just because we can't parse a certain record, doesn't mean we can't try to parse the next record. If a user wants to read "all records the driver can possible understand" they'd have to first know which those are and filter them out somehow with cypher if the driver didn't provide a way to continue reading the record stream after such a hickup.

thelonelyvulpes: I agree with @injectives on this point. It should fail (ideally the query would fail at the server side) similar to as if you were to try and create a 65 bit integer in the database.

injectives: Java driver currently follows the logic of failing when it has problems with unpacking types. This is not exclusive to datetime. If we want to change this, we should probably discuss it as a unification thing.

robsdedude: Are we on the same page @thelonelyvulpes ? I'm talking about the scenario where the driver receives a zone id from the server that's unknown to the driver. What could the server possibly do about this or even know about this in advance?

injectives: In my view, if `Result` throws on record consumption, it should no longer be considered to be in a valid state. One option that we could consider is to throw on accessing the actual value.

injectives: In my view, if `Result` throws on record streaming, it should no longer be seen as a valid handle. One option that we could consider is to throw on accessing the actual value.

injectives: So, in case of Java it would throw on line 4:
```
Result result = session.run("SOMETHING");
Record record = result.next();
Value value = record.get("date");
ZonedDateTime time = value.asZonedDateTime(); // throws here 
```

robsdedude: Summary of off-GH discussion: for 4.4 it'll be what ever is non-breaking. For instance, for Java this will mean break the Result, for Python keep it usable. For 5.0 it should keep the Result object in tact. Throwing when actually accessing the property/attribute in question is a good option here.

PR #490, opened 2022-07-05
Expect 5.0 nightly server to not have UTC fix

robsdedude:
The 5.0 server will take a little longer to ship the UTC fix in BOLT v5.
Instead of skipping those tests, I decided to expect them to fail so that we'll
notice as soon as the fix is in. Else, the tests might remain skipped for too
long.


thelonelyvulpes (approved):
super useful.

bigmontz (approved):
🚗

PR #496, opened 2022-07-19
Make stub server fail hard on OS errors

fbiville:
Signed-off-by: Florent Biville <florent.biville@neo4j.com>


robsdedude (approved):
:zap:

robsdedude:
Seems like some tests make the assumption that loosing connectivity in the middle of the script is fine :detective:

PR #498, opened 2022-07-25
Use safe integers for JSON serialization

fbiville:


robsdedude:
I'm against not testing the corner cases. I want those numbers to be tested! Maybe JSON wasn't the right choice or what, but just leaving the corner areas of the well defined value domain uncovered seems like the wrong approach to the problem.

fbiville:
Closing this in favor of an internal ticket to replace JSON as an exchange format.

PR #501, opened 2022-07-28
Add xml reporting to test suites

bigmontz:
Co-authored-by: Florent Biville <florent.biville@neo4j.com>

### DIFF
--- a/driver.py
+++ b/driver.py
@@ -113,6 +113,7 @@ class Container:
                     config: neo4j.Config):
         env = self._default_env()
         env.update({
+            "TEST_CONFIG_NAME":   config.name,
             "TEST_NEO4J_HOST":    hostname,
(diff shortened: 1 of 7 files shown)


robsdedude:
Is there still a need for XML output?

fbiville:
It _might_ make the output easier to digest by TeamCity since JUnit XML format is quite well supported in general.

robsdedude:
I think https://github.com/neo4j-drivers/testkit/pull/553 might be good enough:tm:  for now? If you disagree, please resolve the conflicts. I'd like to see what the output looks like in TC.

Also, I'd like the XML runner only to be chose in TC builds. For local development I really don't want to have to juggle around XML. Plain text on my command promt is very much appreciated.

fbiville:
Let's close this for now

PR #504, opened 2022-08-03
Fix connection acquisition timeout

robsdedude:
When the driver is performing routing, the connection acquisition for a
connection to the router should also be affected by the connection acquisition
timeout. However, it should have its own timeout because it's a separate
acquisition.

Depends on:
 - https://github.com/neo4j/neo4j-go-driver/pull/385
 - https://github.com/neo4j/neo4j-java-driver/pull/1281

### DIFF
--- a/tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
+++ b/tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
@@ -1,5 +1,3 @@
-import pytest
-
 from nutkit.frontend import Driver
 import nutkit.protocol as types
 from tests.shared import (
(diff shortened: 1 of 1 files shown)


tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
```diff
@@ -178,15 +180,15 @@ def test_should_fail_when_connection_timeout_is_reached_first(self):
         with self.assertRaises(types.DriverError):
             list(self._session.run("RETURN 1 AS n"))
 
-    def test_does_not_encompass_router_handshake(self):
+    def test_router_handshake_has_own_timeout_in_time(self):
         self._start_server(self._router, "router_hello_delay.script")
-        self._start_server(self._server, "session_run.script")
+        self._start_server(self._server, "session_run_auth_delay.script")
```
bigmontz, line 183: Why did you need to add delay to the session run?

bigmontz (approved):
⏲️ 💣

tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
robsdedude: Because I want to make sure that the connection_acquisition_timeout_ms (`6000ms`) counts for each connection acquisition separately. So even though the total delay (4s during router handshake, 4s during reader handshake) exceeds the set timeout, it should still work because the individual delays are below the timeout.

fbiville (approved):

injectives (approved):

PR #508, opened 2022-08-09
Update TestNeo4jBookmarkManager to list results

injectives:
Explicit listing is needed for those drivers that may not necessarily consume all result and may do a `DISCARD` instead. For instance, reactive.


bigmontz (approved):
🐰

PR #509, opened 2022-08-24
Moving BookmarkManager configuration to `driver.session`

bigmontz:


nutkit/frontend/driver.py
```diff
@@ -128,16 +131,20 @@ def close(self):
...
             impersonatedUser=impersonated_user,
-            ignore_bookmark_manager=ignore_bookmark_manager
+            bookmark_manager=bookmark_manager
         )
         res = self.send_and_receive(req, allow_resolution=False)
         if not isinstance(res, protocol.Session):
             raise Exception("Should be session")
+
+        if bookmark_manager is not None:
+            self._bookmarks_managers[bookmark_manager.id] = bookmark_manager
```
injectives: As `BookmarksSupplierRequest` might come during session creation, would it be possible to move this block to line 141 please?

bigmontz: Sure.

boltstub/parsing.py
```diff
@@ -184,7 +184,7 @@ def update_context(self, ctx: "ScriptContext"):
             )
 
 
-class MessageLine(Line, abc.ABC):
+class MessageLine(Line):
```
robsdedude: Can be undone, moved to https://github.com/neo4j-drivers/testkit/pull/510

boltstub/simple_jolt/common/types.py
```diff
@@ -1,4 +1,3 @@
-import abc
```
robsdedude: Can be undone, moved to https://github.com/neo4j-drivers/testkit/pull/510

boltstub/simple_jolt/common/types.py
```diff
@@ -19,7 +18,7 @@ def __init__(self, types):
         self.types = types
 
 
-class _JoltParsedType(JoltType, abc.ABC):
```
robsdedude: Can be undone, moved to https://github.com/neo4j-drivers/testkit/pull/510

nutkit/frontend/driver.py
```diff
@@ -1,8 +1,8 @@
 from typing import Optional
 
+from nutkit.protocol.responses import BookmarkManager
```
```suggestion
```

nutkit/frontend/driver.py
```diff
@@ -128,16 +131,20 @@ def close(self):
 
     def session(self, access_mode, bookmarks=None, database=None,
                 fetch_size=None, impersonated_user=None,
-                ignore_bookmark_manager=None):
+                bookmark_manager: Optional[BookmarkManager] = None):
```
```suggestion
                bookmark_manager=None):
```
I'm against type hinting only half of method signature and we still don't have a checker in place that could've flagged this or any mistake, really.

nutkit/protocol/requests.py
```diff
@@ -214,8 +212,19 @@ class BookmarksConsumerCompleted:
     Signal the method call has finished
     """
 
-    def __init__(self, requestId):
+    def __init__(self, requestId, bookmarkManagerId):
         self.requestId = requestId
+        self.bookmarkManagerId = bookmarkManagerId
+
+
+class NewBookmarkManager:
```
robsdedude, line 218: Should we have a method to close/destroy BMMs to avoid piling up resources in the backend?

nutkit/frontend/driver.py
```diff
@@ -13,15 +13,12 @@ def __init__(self, backend, uri, auth_token, user_agent=None,
...
-                 bookmark_manager_config: Optional[BMMConfig] = None):
+                 connection_acquisition_timeout_ms=None):
         self._backend = backend
         self._resolver_fn = resolver_fn
         self._domain_name_resolver_fn = domain_name_resolver_fn
-        self._bookmark_manager_config = bookmark_manager_config
-        bookmark_manager = from_bookmark_manager_config_to_protocol(
-            bookmark_manager_config
-        )
+        self._bookmarks_managers = {}
```
robsdedude: Why are the BMMs bound to a driver object? Feels like they should be a global list/dict somewhere. Or in their own class.

nutkit/protocol/responses.py
```diff
@@ -81,9 +88,10 @@ class BookmarksSupplierRequest:
     bookmarks for all databases.
     """
 
-    def __init__(self, id, database=None):
+    def __init__(self, id, bookmarkManagerId, database=None):
         # id of the callback request
         self.id = id
+        self.bookmarkManagerId = bookmarkManagerId
```
robsdedude:
```suggestion
        self.bookmark_manager_id = bookmarkManagerId
```

I'm ok with breaking Python's style guide for the things that go over the wire. That's why the request classes have attributes that are camel case instead of snake case. Hence, I'm ok with the response classes having camel case parameters. But right after that one layer, I'm voting for re-writing it to be PEP8 conform.

nutkit/protocol/responses.py
```diff
@@ -95,9 +103,10 @@ class BookmarksConsumerRequest:
     to send the new bookmark set for a given database.
     """
 
-    def __init__(self, id, database, bookmarks):
+    def __init__(self, id, bookmarkManagerId, database, bookmarks):
         # id of the callback request
         self.id = id
+        self.bookmarkManagerId = bookmarkManagerId
```
robsdedude:
```suggestion
        self.bookmark_manager_id = bookmarkManagerId
```

tests/stub/driver_parameters/test_bookmark_manager.py
```diff
@@ -280,17 +290,27 @@ def test_should_use_initial_bookmark_set_in_the_fist_tx(self):
         self._start_server(self._router, "router_with_db_name.script")
         self._start_server(self._server, "transaction_chaining.script")
 
-        self._driver = self._new_driver(Neo4jBookmarkManagerConfig(
-            initial_bookmarks={"neo4j": ["fist_bm"]}
-        ))
+        (self._driver, manager) = self._new_driver_and_bookmark_manager(
```
robsdedude:
```suggestion
        self._driver, manager = self._new_driver_and_bookmark_manager(
```

tests/stub/driver_parameters/test_bookmark_manager.py
```diff
@@ -313,17 +333,27 @@ def test_should_send_all_db_bookmarks_and_update_only_relevant(self):
         self._start_server(self._router, "router_with_db_name.script")
         self._start_server(self._server, "transaction_chaining.script")
 
-        self._driver = self._new_driver(Neo4jBookmarkManagerConfig(
-            initial_bookmarks={"neo4j": ["fist_bm"], "adb": ["adb:bm1"]}
-        ))
+        (self._driver, manager) = self._new_driver_and_bookmark_manager(
```
robsdedude:
```suggestion
        self._driver, manager = self._new_driver_and_bookmark_manager(
```

tests/stub/driver_parameters/test_bookmark_manager.py
```diff
@@ -436,17 +484,19 @@ def test_should_resolve_database_name_with_system_bookmarks(self):
         self._start_server(self._router, "router_with_db_name.script")
         self._start_server(self._server, "transaction_chaining.script")
 
-        self._driver = self._new_driver(Neo4jBookmarkManagerConfig(
-            initial_bookmarks={"system": ["sys:bm1"]}
-        ))
+        (self._driver, manager) = self._new_driver_and_bookmark_manager(
```
robsdedude:
```suggestion
        self._driver, manager = self._new_driver_and_bookmark_manager(
```

tests/stub/driver_parameters/test_bookmark_manager.py
```diff
@@ -494,7 +532,7 @@ def get_bookmarks(db):
             get_bookmarks_calls.append([db])
             return []
 
-        self._driver = self._new_driver(
+        (self._driver, manager) = self._new_driver_and_bookmark_manager(
```
robsdedude:
```suggestion
        self._driver, manager = self._new_driver_and_bookmark_manager(
```

tests/stub/driver_parameters/test_bookmark_manager.py
```diff
@@ -558,7 +600,7 @@ def get_bookmarks(db):
                 return system_bookmarks + neo4j_bookmarks
             return bookmarks.get(db, [])
 
-        self._driver = self._new_driver(
+        (self._driver, manager) = self._new_driver_and_bookmark_manager(
```
robsdedude:
```suggestion
        self._driver, manager = self._new_driver_and_bookmark_manager(
```

tests/stub/driver_parameters/test_bookmark_manager.py
```diff
@@ -603,7 +649,7 @@ def test_should_call_bookmarks_consumer_when_new_bookmarks_arrive(self):
         def bookmarks_consumer(db, bookmarks):
             bookmarks_consumer_calls.append([db, bookmarks])
 
-        self._driver = self._new_driver(
+        (self._driver, manager) = self._new_driver_and_bookmark_manager(
```
robsdedude:
```suggestion
        self._driver, manager = self._new_driver_and_bookmark_manager(
```

tests/stub/driver_parameters/test_bookmark_manager.py
```diff
@@ -733,3 +846,26 @@ def supports_minimal_bookmarks(self):
...
+            bookmark_manager_config = Neo4jBookmarkManagerConfig()
+        return create_bookmark_manager(
+            self._backend,
+            bookmark_manager_config
+        )
+
+    def _new_driver_and_bookmark_manager(self, bookmark_manager_config=None):
+        bookmark_manager = self._new_bookmark_manager(bookmark_manager_config)
+        driver = self._new_driver()
+        return (driver, bookmark_manager)
```
robsdedude:
```suggestion
        return driver, bookmark_manager
```

nutkit/frontend/driver.py
bigmontz: I kept here because it makes the object more restricted to a scope and it was the more restricted scope I figured out to have it registered and still be able of trigger the extension functions.

The Bookmark Managers still be able to be created as a standalone object, but when associated to the session, the driver object will keep the reference for triggering the callbacks if needed.

About keep the references as global, Driver objects (tk side) accessing the same list of BMM can cause really complicated to debug issues. I prefer not. And the ideal scenario, the Session will hold the reference. However, I don't think it worth the re-factory for now.

nutkit/frontend/driver.py
bigmontz: My auto-complete is sad now, but I will accept the suggestion.

nutkit/protocol/requests.py
```diff
@@ -214,8 +212,19 @@ class BookmarksConsumerCompleted:
     Signal the method call has finished
     """
 
-    def __init__(self, requestId):
+    def __init__(self, requestId, bookmarkManagerId):
```
robsdedude: No need to send the bmm id. The backend just initiated this callback. It shouldn't be too hard for it to remember which BMM it was that triggered the callback. So I vote for having only a single source of truth (i.e., `requestId`). (Same comment for `BookmarksSupplierCompleted`).

nutkit/frontend/driver.py
```diff
@@ -59,29 +55,36 @@ def receive(self, timeout=None, hooks=None, *, allow_resolution):
                         hooks=hooks
                     )
                     continue
-            if self._bookmark_manager_config is not None:
-                supply = self._bookmark_manager_config.bookmarks_supplier
-                if supply is not None:
-                    if isinstance(res, protocol.BookmarksSupplierRequest):
+            if isinstance(res, protocol.BookmarksSupplierRequest):
+                if res.bookmarkManagerId in self._bookmarks_managers:
```
robsdedude: Shouldn't we fail hard if the backend requests a callback on an undefined BMM?

nutkit/protocol/requests.py
bigmontz: We will only have 20 lightweight objects in the backend. The log messages generated for this call in the JS backend concern me more. I already removed some log messages from the backend for resolving issues with the process being killed.

nutkit/frontend/driver.py
robsdedude: It was the wrong type anyway :grimacing: It was a protocol BMM but should've been a frontend BMM...

nutkit/frontend/driver.py
bigmontz: It will fail, the res will be return and the receiver will get an expected object and fail. This is the behaviour with the resolution messages. We could raise for being more explicit.

robsdedude (approved):

nutkit/frontend/bookmark_manager.py
```diff
@@ -20,22 +24,60 @@ class Neo4jBookmarkManagerConfig:
 
 @dataclass
 class BookmarkManager:
```
bigmontz, line 26: BookmarkManager is in charge of:

* Keep track of the bookmark manager registry
* Process callback by select from the registry and execute the logic. 
* And be the bookmark manager. 

Maybe we could extract the static part in a class called `BookmarkManagerRegistry` which will not be static. This also enable us to add a method close the registry, this method closes all the bookmark managers created by the registry. The problem going this direction is that we will need to inject the bookmark manager registry in the frontend.driver creation.

nutkit/protocol/requests.py
```diff
@@ -228,6 +228,16 @@ def __init__(self, initial_bookmarks,
         self.bookmarksConsumerRegistered = bookmarks_consumer_registered
 
 
+class BookmarkManagerClose:
+    """Destroy the bookmark manager in the backend and free the resources.
+
+    Backend should respond with a BookmarkManager response.
+    """
```
bigmontz: It may worth emphasise this method is not a part of the BookmarkManager implementation, it is only a helper method for testkit resource management.

nutkit/frontend/driver.py
```diff
@@ -13,15 +10,12 @@ def __init__(self, backend, uri, auth_token, user_agent=None,
...
-                 bookmark_manager_config: Optional[BMMConfig] = None):
+                 connection_acquisition_timeout_ms=None):
         self._backend = backend
         self._resolver_fn = resolver_fn
         self._domain_name_resolver_fn = domain_name_resolver_fn
-        self._bookmark_manager_config = bookmark_manager_config
-        bookmark_manager = from_bookmark_manager_config_to_protocol(
-            bookmark_manager_config
-        )
+        self._bookmarks_managers = {}
```
bigmontz: Leftover

nutkit/frontend/bookmark_manager.py
robsdedude: I agree that the concerns should be separated. I wish I also could refactor the `ResolverResolutionRequired` and `DomainNameResolutionRequired` registries into their own class. But that wouldn't work because they don't send the driverId they were registered with :disappointed:

PR #510, opened 2022-08-25
Fix linter issues by removing un-necessary ABC

bigmontz:
Classes should not be created as abstract if they don't have abstract methods.


robsdedude (approved):
https://github.com/PyCQA/flake8-bugbear/issues/273

PR #511, opened 2022-08-25
unify after run expectations for test_summary stub scripts

thelonelyvulpes:
Not all drivers pipeline a pull after session.run.
Some of test_summary's scripts expect a discard instead of a pull, so unify that.
.NET's Rx driver for example.


robsdedude (approved):
:ballot_box:

robsdedude:
What's the state of this PR?

PR #513, opened 2022-09-02
Fix DateTime test

robsdedude:
Test started failing because 4.4 servers with the UTC patch started appearing.
The test did not account for all implications of the new protocol specs.

### DIFF
--- a/tests/neo4j/datatypes/test_temporal_types.py
+++ b/tests/neo4j/datatypes/test_temporal_types.py
@@ -180,6 +180,26 @@ class TestDataTypes(_TestTypesBase):
                         self._verify_can_echo(dt)
 
     def test_date_time_cypher_created_tz_id(self):
+        def assert_utc_equal(dt_, cypher_dt_):
+            self.assertEqual(dt_.timezone_id, tz)
+            # We are comparing in UTC because the server's and the
+            # driver's timezone db may diverge.
+            self.assertEqual(dt_.as_utc(), cypher_dt_.as_utc())
+
+        def assert_wall_time_equal(dt_, cypher_dt_):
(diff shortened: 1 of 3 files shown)


tests/neo4j/datatypes/test_temporal_types.py
```diff
@@ -204,24 +224,37 @@ def work(tx):
             return map(lambda x: getattr(x, "value", x), rec.values)
 
         self._create_driver_and_session()
+        server_supports_utc = get_server_info().has_utc_patch
```
fbiville, line 227: why is it not a simple boolean?

fbiville: ok, I found out why below.
maybe it could be renamed `server_version_may_support_utc` :P

fbiville (approved):

tests/neo4j/datatypes/test_temporal_types.py
robsdedude: Yeah, for 4.4 and 4.3 there's no straight-forward way for TestKit to know if the server supports the UTC patch or not. So this value can be `1` yes, `0` no, and `0.5` maybe :grimacing:

bigmontz (approved):
📅

tests/neo4j/datatypes/test_temporal_types.py
bigmontz: trilean.

thelonelyvulpes: might we consider an enum?

robsdedude: https://github.com/neo4j-drivers/testkit/pull/515

PR #518, opened 2022-09-07
Add test case where `ROUTE` returns `FAILURE` with code equals to `Neo.TransientError.General.DatabaseUnavailable`

bigmontz:
This test should check if this will make the router be forgotten and the new routing table be acquired


tests/stub/routing/scripts/v5x0/router_yielding_database_unavailable_failure.script
```diff
@@ -0,0 +1,8 @@
+!: BOLT #VERSION#
+!: AUTO RESET
+
+C: HELLO {"scheme": "basic", "credentials": "c", "principal": "p", "user_agent": "007", "routing": #ROUTINGCTX#, "[patch_bolt]": "*"}
+S: SUCCESS {"server": "#SERVER_AGENT#", "connection_id": "bolt-123456789"}
+C: ROUTE #ROUTINGCTX# [] {"[db]": {"U": "*"}}
+S: FAILURE {"code": "Neo.ClientError.General.Unknown", "message": "wut!"}
```
fbiville: the error does not match what the filename describes, or what am I missing?

bigmontz: I forgot to the change the error code

tests/stub/routing/test_routing_v5x0.py
```diff
@@ -1873,6 +1873,59 @@ def work(tx):
         self.assertEqual([[]], sequences)
         self.assertEqual(2, try_count)
 
+    def test_should_forget_router_address_on_database_unavailable_error(self):
```
robsdedude, line 1876: I don't understand what this test asserts and how it flows.

* Driver contacts initial router that gives `DatabaseUnavailable`.
* Driver retries (on initial router again) and encounters router that works.
* Driver contacts writer2 that's unavailable.
* Assert no result was fetched.

Here's what I don't understand:
* How is `self._writeServer1` ever being contacted? (After all `self._writeServer1.done()` passes)
* How does this flow assert that any router was removed from the routing table? The driver just retries the initial router. I believe this test only asserts that the given error code is being retried.

bigmontz: Flow:

> Driver contacts the initial router, get a routing table point to writer 1 and router 2
> Driver tries to run the tx in writer 1, get a Db unavailable error and remove the writer from the list
> Driver doesn't have any writer, so it contacts router 2 for getting the routing table and get another db unavailable 
> Driver contacts the initial router, which now return writer 2 as the writer
> Driver contacts writer 2 for running the query and it gets no records from it, but that's fine.

fbiville: > which now return writer 2 as the writer

this is implemented via a hook, but could be scripted instead (I think), as we discussed with Antonio before

fbiville: actually, the hook is more accurately applied than the script, so I like the idea more the more I think about it now
`</monologue>`

tests/stub/routing/test_routing_v5x0.py
```diff
@@ -1873,6 +1873,59 @@ def work(tx):
...
+
+        def work(tx):
+            nonlocal try_count
+            try_count = try_count + 1
+            result = tx.run("RETURN 1 as n")
+            sequences.append(self.collect_records(result))
+
+        retried = False
+
+        def on_retryable_negative_hook(_):
```
robsdedude: Nit
```suggestion
        def retryable_negative_hook(_):
```
or

```suggestion
        def on_retryable_negative(_):
```

tests/stub/routing/test_routing_v5x0.py
```diff
@@ -1873,6 +1873,59 @@ def work(tx):
...
+            nonlocal retried
+            if not retried:
+                self._routingServer1.done()
```
robsdedude, line 1909: You could probably also assert that `self._writeServer1.done()` here already.

robsdedude (approved):
I just looked at `writer_tx_yielding_database_unavailable_failure.script` and realized that it returns the error on `PULL`. I'm not sure a real server would do that, but that shouldn't be part of this PR either way. :heavy_check_mark:

tests/stub/routing/test_routing_v5x0.py
robsdedude: Thanks. I missed the point that the initial router is actually a good one.

tests/stub/routing/test_routing_v5x0.py
```diff
@@ -1873,6 +1873,59 @@ def work(tx):
...
+                self.start_server(
+                    self._routingServer1,
+                    "router_yielding_writer2.script"
+                )
+            retried = True
+
+        session.write_transaction(work, hooks={
+            "on_send_RetryableNegative": on_retryable_negative_hook
+        })
```
bigmontz:
```suggestion
        session.write_transaction(work, hooks={
            "on_send_RetryableNegative": on_retryable_negative
        })
```

tests/stub/routing/test_routing_v5x0.py
bigmontz: I don't know if the driver already release the connection back to the pool at this point in time

PR #521, opened 2022-09-13
[4.4] Test session close with pending auto-commit causing error

robsdedude:
:cherries: :pick: of https://github.com/neo4j-drivers/testkit/pull/522


robsdedude:
Depending on
 * https://github.com/neo4j-drivers/testkit/pull/523 
 * https://github.com/neo4j/neo4j-python-driver/pull/802
 * https://github.com/neo4j/neo4j-java-driver/pull/1310

robsdedude:
I don't think this work will be completed. It's a breaking change that needs coordination between drivers. Maybe we'll come back to it should it every become an actual problem.

PR #522, opened 2022-09-13
Test session close with pending auto-commit causing error

robsdedude:


robsdedude:
Depending on
 * https://github.com/neo4j/neo4j-python-driver/pull/800
 * https://github.com/neo4j/neo4j-java-driver/pull/1311

tests/stub/session_run/scripts/session_error_on_discard_result.script
```diff
@@ -0,0 +1,16 @@
...
+*: RESET
+C: RUN {"U": "*"} {"{}": "*"} {"{}": "*"}
+S: SUCCESS {"fields": ["n"]}
+{*
+    C: PULL {"n": {"Z": "*"}, "[qid]": -1}
+    S: RECORD [1]
+       RECORD [2]
+       SUCCESS {"type": "r", "has_more": true}
+*}
+C: DISCARD {"n": -1, "[qid]": -1}
```
bigmontz, line 13: The problem with this is the case where the `autocommit tx` is waiting for a lock, it will still wait until it times out or acquire the lock, instead of release everything on the `session.close` by calling `RESET`

robsdedude: 1) Currently the Python, Java, and .NET drivers will wait for the auto-commit tx to be finished. If find this a sane option (see my arguments on Slack, that I don't want to repeat here)
  2) I agree that having a "cancel that session" (sending a `RESET`) API is handy to have. But close should not be that API, I think.
  
  I think this calls for a unification meeting.

robsdedude:
I don't think this work will be completed. It's a breaking change that needs coordination between drivers. Maybe we'll come back to it should it every become an actual problem.

PR #524, opened 2022-09-14
Save disk space when `TEST_DOCKER_RMI` is used

robsdedude:
* Load all TC nightly images directly from their TC build. This allows us to get rid of the build that just collected the docker loadable tars. This should be easier to maintain. Said build started to become too big as it was collecting to many artifacts. It was frankly also just a waste of storage.
 * `docker rmi` neo4j images whenever switching the test image. This saves disk space (e.g., in CI pipelines) at the cost of not having the images cached.


main.py
```diff
@@ -299,19 +315,18 @@ def _exit():
 
     # Important to stop all docker images upon exit
     # Also make sure that none of those images are running at this point
+    cleanup = build_cleanup(settings)
     atexit.register(cleanup)
     cleanup()
```
bigmontz, line 311: Why do we call cleanup just after we register it in the atexit? Won't it be called twice?

bigmontz (approved):
🗡️

main.py
robsdedude: I mean that was there before. I believe that's just to clean the environment in case there is some stuff there already. Probably not necessary though.

PR #525, opened 2022-09-14
[4.4] Save disk space when `TEST_DOCKER_RMI` is used

robsdedude:
* Load all TC nightly images directly from their TC build. This allows us to get rid of the build that just collected the docker loadable tars. This should be easier to maintain. Said build started to become too big as it was collecting to many artifacts. It was frankly also just a waste of storage.
 * `docker rmi` neo4j images whenever switching the test image. This saves disk space (e.g., in CI pipelines) at the cost of not having the images cached.

Port of https://github.com/neo4j-drivers/testkit/pull/524


bigmontz (approved):
🗡️

PR #527, opened 2022-09-19
Ignore non-0 exit code from killed processes in docker.

robsdedude:
`docker stop` (or `docker rm -f`) will first send `SIGTERM` to the process started with `docker run` (`docker exec` processes don't receive a signal at all). When not shutting down gracefully, they'll be `SIGKILL`ed. Either way, the exit code of the `docker run` command will be (assuming running on Linux) non-zero https://tldp.org/LDP/abs/html/exitcodes.html. Therefore, we want to ignore any failures (non-zero exits) during shut down of docker containers.


bigmontz (approved):
✊🏼

gjmwoods (approved):

PR #528, opened 2022-10-03
Allow scripts to finish when driver ignores timeouts

robsdedude:
Spotted by @klobuczek

Drivers should be able to successfully finish the scripts if they ignore the timeout hint. This way, they will not raise an error as expected by the tests, hence fail them.


bigmontz (approved):
🔧

PR #529, opened 2022-10-06
Fix docker run commands not logging output

robsdedude:
Broke in https://github.com/neo4j-drivers/testkit/pull/527

### DIFF
--- a/docker.py
+++ b/docker.py
@@ -21,9 +21,11 @@ def _docker_path(path):
 def _subprocess_run(cmd, *args, log_path=None, background=False, **kwargs):
     class Runner:
         def __init__(self):
-            if log_path and "stdout" in kwargs or "stderr" in kwargs:
+            if log_path and ("stdout" in kwargs or "stderr" in kwargs):
                 raise ValueError("Cannot specify log_path with "
                                  "stdout or stderr")
+            kwargs.setdefault("stdout")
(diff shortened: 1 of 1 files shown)


docker.py
```diff
@@ -40,6 +45,7 @@ def run(self):
                 err_path = os.path.join(log_path, "err.log")
                 with open(out_path, "a") as out_fd:
                     with open(err_path, "a") as err_fd:
+                        kwargs.update(stdout=out_fd, stderr=err_fd)
```
thelonelyvulpes, line 48: why do we change kwargs after it is used?

thelonelyvulpes (commented):
makes sense. little unsure about line 48

docker.py
robsdedude: Ah :see_no_evil: you spotted how ugly this is. I didn't feel like making it nice. But now I feel exposed and want to.
For explanation `kwargs` is a dict that bound to the whole scope of the `_subprocess_run` function and `_process_run` uses those kwargs. So I'm modifying the `kwargs` dict in place to change what `_process_run` does. It's as ugly and hard to follow as using global variables :nauseated_face:

thelonelyvulpes (approved):

docker.py
thelonelyvulpes: Makes sense, thank you for the explanation.

PR #531, opened 2022-10-12
Introduce `Driver.execute_query`

bigmontz:


nutkit/protocol/responses.py
```diff
@@ -495,6 +495,18 @@ def __init__(self, inUse, idle):
...
 
+class EagerResult:
+    def __init__(self, keys, records, summary):
+        self.keys = keys
+        record_list = []
+        if records is not None:
+            for record in records:
+                record_list.append(Record(values=record["values"]))
+
+        self.records = record_list
```
robsdedude:
```suggestion
        self.records = [Record(**record) for record in records or []]
```

Difference: 
 * Will pass all elements of `records` to `Record`. But there shouldn't be any extra ones, should there?
 * Shorter and most likely more performant.

tests/stub/driver_execute_query/test_driver_execute_query.py
```diff
@@ -0,0 +1,121 @@
+from nutkit.frontend import Driver
+import nutkit.protocol as types
+from tests.shared import TestkitTestCase
+from tests.stub.shared import StubServer
+
+
+class TestDriverExecuteQuery(TestkitTestCase):
+    required_features = (
+        types.Feature.BOLT_5_0,
+        types.Feature.API_BOOKMARK_MANAGER,
```
robsdedude, line 14: I'm voting for adding a feature flag for this API. I imagine not all community drivers will want to pick it up...

tests/stub/driver_execute_query/test_driver_execute_query.py
```diff
@@ -0,0 +1,121 @@
...
+        self._reader.reset()
+        self._writer.reset()
+        return super().tearDown()
+
+    def test_execute_query_without_params_and_config(self):
+        self._start_server(self._router, "router.script")
+        self._start_server(self._writer, "tx_return_1.script")
+        self._driver = self._new_driver()
+
+        eager_result = self._driver.execute_query("RETURN 1 as n")
```
robsdedude:
```suggestion
        eager_result = self._driver.execute_query("RETURN 1 AS n")
```
\+ everywhere else + in the scripts

tests/stub/driver_execute_query/test_driver_execute_query.py
```diff
@@ -0,0 +1,121 @@
...
+        self.assertIsNotNone(eager_result.summary)
+
+    def test_configure_routing_to_writers(self):
+        self._start_server(self._router, "router.script")
+        self._start_server(self._writer, "tx_return_1_with_params.script")
+        self._driver = self._new_driver()
+
+        eager_result = self._driver.execute_query("RETURN 1 as n", {
+            "a": types.CypherInt(1)
+        }, {"routing": "W"})
```
robsdedude:
```suggestion
        }, {"routing": "w"})
```
I vote for keeping them lower case as that's also the protocol for `NewSession`.

tests/stub/driver_execute_query/test_driver_execute_query.py
bigmontz: Sure, I will add.

nutkit/protocol/feature.py
```diff
@@ -13,6 +13,9 @@ class Feature(Enum):
     # is picked up instead or we need to wait until the full pool depletes.
     API_CONNECTION_ACQUISITION_TIMEOUT = \
         "Feature:API:ConnectionAcquisitionTimeout"
+    # The driver offers a method for run a query in a retriable context at the
+    # driver object level
+    API_DRIVER_EXECUTE_QUERY = "Feature:API:Driver:ExecuteQuery"
```
thelonelyvulpes: it's not the most important change but, `Feature:API:Driver.ExecuteQuery`
GetServerInfo is the exception that proves the rule I guess

robsdedude: Agreed. Shame on me for introducing `Feature:API:Driver:GetServerInfo`. All the others (e.g., `Feature:API:Result.List`, `Feature:API:Driver.VerifyConnectivity`) use dots. Admittedly, there are quite a few that use dots, that shouldn't. Some clean-up may be in order some time later...

tests/stub/driver_execute_query/test_driver_execute_query.py
```diff
@@ -0,0 +1,262 @@
...
+        self._start_server(
+            self._writer, "transaction_chaining.script")
+        self._driver = self._new_driver()
+
+        # CREATING NODE
+        eager_result = self._driver.execute_query(
+            "CREATE (p:Person{name:$name}) RETURN p.name AS name", {
+                "name": types.CypherString("the person")
+            }, {
+                "bookmarkManagerId": None
```
robsdedude: Note that some languages have a difficult time differentiating a missing key from a key with a `null` value in a JSON object.

However, I'm not sure we have to do something about it as there are other parts of the TestKit protocol that also require the backends to be sensitive to this.

Another thing regarding this: I'm not a huge fan of magic strings in the tests. So I find the fact that the tests need to know that this config option is called `"bookmarkManagerId"` not ideal. If you want to, I can propose an alternative design.

bigmontz: I was thinking about using `**kwargs` instead and let the translation to the protocol message, specially because the "bookmarkManagerId" has difference between not bet set or be set to None. Maybe we could work around using `-1` for when it is `not set`

robsdedude: `-1`: "use no BMM", `null`: "use default BMM" sounds like a good idea to me.

Instead of using `**kwargs`, you could use something like

```python
def execute_query(self, cypher, params=None, database=None, other_config_option_1=None,  other_config_option_2=None, bookmark_manager=...):
    config = {}
    if bookmark_manager is None:
        config["bookmarkManagerId"] = -1
    elif bookmark_manager is not Ellipsis:
        config["bookmarkManagerId"] = bookmark_manager.id
   ...     
```

tests/stub/driver_execute_query/test_driver_execute_query.py
```diff
@@ -0,0 +1,262 @@
...
+    def test_configure_routing_to_writers(self):
+        self._start_server(self._router, "router.script")
+        self._start_server(self._writer, "tx_return_1_with_params.script")
+        self._driver = self._new_driver()
+
+        eager_result = self._driver.execute_query("RETURN 1 AS n", {
+            "a": types.CypherInt(1)
+        }, {"routing": "w"})
+
+        self.assertEqual(eager_result.keys, ["n"])
```
thelonelyvulpes: Can we fix all the assertions being backwards, I know it's a common issue in testkit but it should be (expected, actual). I'm happy to do it

robsdedude: That's documented nowhere. The order is arbitrary. It's just PyCharm that felt entitled to assign the first argument as the expected value.

https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertEqual

thelonelyvulpes: unittest truly is the worst why don't they have this defined. It is my understanding that in most assertion libraries you assert(expect, actual)

bigmontz: It was my js doing `expect(this).toEqual(that)`

tests/stub/driver_execute_query/test_driver_execute_query.py
bigmontz: Sure, I will address it.

tests/stub/driver_execute_query/test_driver_execute_query.py
robsdedude: Just ignore PyCharm it's stupid. Here goes PyCharm's output with pytest
```python
def test_foo(self):
    assert 42 != 42
```
leads to
```
42 != 42

Expected :42
Actual   :42
```
:face_with_head_bandage:

nutkit/frontend/driver.py
```diff
@@ -67,7 +67,22 @@ def send_and_receive(self, req, timeout=None, hooks=None, *,
...
+        if (routing):
+            config["routing"] = routing
+        if (database):
+            config["database"] = database
+        if (impersonated_user):
+            config["impersonatedUser"] = impersonated_user
+
+        if bookmark_manager is None:
+            config["bookmarkManagerId"] = -1
+        elif bookmark_manager is not Ellipsis:
```
fbiville, line 81: what's `Ellipsis`? 🤔

tests/stub/driver_execute_query/scripts/transaction_chaining.script
```diff
@@ -0,0 +1,38 @@
...
+!: ALLOW CONCURRENT
+
+A: HELLO {"{}": "*"}
+*: RESET
+{*
+    {{
+        C: BEGIN {"[db]": "*"}
+        S: SUCCESS {}
+        {{
+            C: RUN "CREATE (p:Person{name:$name}) RETURN p.name AS name" { "name":"the person" } {}
```
fbiville: can we make the query simpler? does it need to be that long? (applies to other places too)

tests/stub/driver_execute_query/test_driver_execute_query.py
```diff
@@ -0,0 +1,256 @@
...
+        self._start_server(self._router, "router.script")
+        self._start_server(self._writer, "tx_return_1.script")
+        self._driver = self._new_driver()
+
+        eager_result = self._driver.execute_query("RETURN 1 AS n")
+
+        self.assertEqual(eager_result.keys, ["n"])
+        self.assertEqual(eager_result.records, [
+                         types.Record(values=[types.CypherInt(1)])])
+        self.assertIsNotNone(eager_result.summary)
```
fbiville: you could check the query included the summary

tests/stub/driver_execute_query/test_driver_execute_query.py
```diff
@@ -0,0 +1,256 @@
...
+
+    def test_execute_query_without_config(self):
+        self._start_server(self._router, "router.script")
+        self._start_server(self._writer, "tx_return_1_with_params.script")
+        self._driver = self._new_driver()
+
+        eager_result = self._driver.execute_query("RETURN 1 AS n", {
```
fbiville:
```suggestion
        eager_result = self._driver.execute_query("RETURN $a AS n", {
```

would be more logical (applies to other places too)

tests/stub/driver_execute_query/test_driver_execute_query.py
```diff
@@ -0,0 +1,256 @@
...
+            "a": types.CypherInt(1)
+        }, database="neo4j")
+
+        self.assertEqual(eager_result.keys, ["n"])
+        self.assertEqual(eager_result.records, [
+                         types.Record(values=[types.CypherInt(1)])])
+        self.assertIsNotNone(eager_result.summary)
```
fbiville: you can check the database here

fbiville (commented):
comprehensive suite of tests, nice!

nutkit/frontend/driver.py
bigmontz: the `...`

tests/stub/driver_execute_query/scripts/transaction_chaining.script
bigmontz: I made this way for making it easier to figure out what's going on.

nutkit/protocol/requests.py
```diff
@@ -592,3 +592,19 @@ def __init__(self, result_id, record_key, type_name, field_id):
         self.recordKey = record_key
         self.type = type_name
         self.field = field_id
+
+
+class ExecuteQuery:
+    """
+    Request to execute a query in a retriable context.
+
+    Backend should return EagerResult or a Error response.
```
robsdedude, line 601: Is it worth documenting the fields, their types, and meaning? Especially the `config.bookmarkManagerId` values that have a special meaning.

tests/stub/driver_execute_query/test_driver_execute_query.py
bigmontz: I checked some fields, the summary is quite extensive. Maybe we could adapt the `test_summary` for testing the summary returned in the eager function.

nutkit/protocol/requests.py
bigmontz: I've put some docs in a semi structured way.

tests/stub/driver_execute_query/test_driver_execute_query.py
bigmontz: I double check the result of a failure in the terminal and it is getting like this:

```
Traceback (most recent call last):
  File "./tests/stub/driver_execute_query/test_driver_execute_query.py", line 261, in test_configure_custom_bookmark_manager
    self.assertEqual(summary2.query_type, "w")
AssertionError: 'r' != 'w'
- r
+ w
```

I will invert everything here

nutkit/protocol/feature.py
```diff
@@ -13,6 +13,9 @@ class Feature(Enum):
     # is picked up instead or we need to wait until the full pool depletes.
     API_CONNECTION_ACQUISITION_TIMEOUT = \
         "Feature:API:ConnectionAcquisitionTimeout"
+    # The driver offers a method for run a query in a retriable context at the
```
robsdedude:
```suggestion
    # The driver offers a method for run a query in a retryable context at the
```
I vote for retryable (also in test names, etc.) As that's the spelling more commonly used throughout TestKit.

nutkit/frontend/driver.py
```diff
@@ -67,6 +67,28 @@ def send_and_receive(self, req, timeout=None, hooks=None, *,
...
+    def execute_query(self, cypher, params=None, config=None,
+                      routing=None, database=None,
+                      impersonated_user=None, bookmark_manager=...):
+        config = {}
+        if (routing):
+            config["routing"] = routing
+        if (database):
+            config["database"] = database
+        if (impersonated_user):
+            config["impersonatedUser"] = impersonated_user
```
```suggestion
        if routing:
            config["routing"] = routing
        if database:
            config["database"] = database
        if impersonated_user:
            config["impersonatedUser"] = impersonated_user
```
This isn't JavaScript :wink: 

But why do we even omit them? `SessionRun` e.g., always transmits `params`.

robsdedude: So what about 
```python
        config = {
            "routing": routing,
            "database": database,
            "impersonatedUser": impersonated_user,
        }
```
instead?
Similarly in `ExecuteQuery`.

nutkit/frontend/driver.py
```diff
@@ -67,6 +67,28 @@ def send_and_receive(self, req, timeout=None, hooks=None, *,
         return self.receive(timeout=timeout, hooks=hooks,
                             allow_resolution=allow_resolution)
 
+    def execute_query(self, cypher, params=None, config=None,
```
```suggestion
    def execute_query(self, cypher, params=None,
```
unused parameter, I believe.

nutkit/protocol/requests.py
```diff
@@ -592,3 +592,31 @@ def __init__(self, result_id, record_key, type_name, field_id):
...
+    :param config.routing: The type of routing ("w" for Writers,
+         "r" for "Readers")
+    :param config.impersonatedUser: The user which will be impersonated
+    :param config.bookmarkManagerId: The id of the bookmark manager
+        used in the query. None or not define for using the default,
+        -1 for disabling the BookmarkManager
+    """
+
+    def __init__(self, driverId, cypher, params, config):
+        self.driverId = driverId
```
robsdedude, line 617:
```suggestion
    def __init__(self, driver_id, cypher, params, config):
        self.driverId = driver_id
```

```suggestion
    :param driver_id: The id of the driver where the cypher query has to run.
    :param cypher: The cypher query which to run.
    :param params: The cypher query params.
    :param config: The configuration
    :param config.database: The database where the query will run.
    :param config.routing: The type of routing ("w" for Writers,
         "r" for "Readers")
    :param config.impersonatedUser: The user which will be impersonated
    :param config.bookmarkManagerId: The id of the bookmark manager
        used in the query. None or not define for using the default,
        -1 for disabling the BookmarkManager
    """

    def __init__(self, driver_id, cypher, params, config):
        self.driverId = driver_id
```

nutkit/frontend/driver.py
bigmontz: This is pythonscript.

tests/stub/driver_execute_query/scripts/transaction_chaining.script
```diff
@@ -0,0 +1,38 @@
...
+               SUCCESS {"type": "w"}
+        ----
+            C: RUN "MATCH (p:Person{name:$name}) RETURN p.name AS name" { "name":"the person" } {}
+            S: SUCCESS {"fields": ["name"]}
+            C: PULL {"n": 1000}
+            S: SUCCESS {"type": "r"}
+        }}
+        C: COMMIT
+        S: SUCCESS {"bookmark": "bm1"}
+    ----
```
robsdedude: Why are we using an alternative inside a loop here? Don't we expect a query with the initial bookmark, then a query with the next bookmark and then no further query?

**EDIT**: Don't we expect a query with the initial bookmark, then a read query with, depending on the test case, either the initial bookmarks or the one from the first query, and then no further queries?

robsdedude: Here my suggestion

```
!: BOLT 5.0

A: HELLO {"{}": "*"}
*: RESET

C: BEGIN {"[db]": "*"}
S: SUCCESS {}
C: RUN "CREATE (p:Person {name:$name}) RETURN p.name AS name" {"name": "the person"} {}
S: SUCCESS {"fields": ["name"]}
C: PULL {"n": 1000}
S: RECORD ["the person"]
   SUCCESS {"type": "w"}
C: COMMIT
S: SUCCESS {"bookmark": "bm1"}
*: RESET

{{
    C: BEGIN {"[db]": "*"}
    S: SUCCESS {}
    C: RUN "MATCH (p:Person {name:$name}) RETURN p.name AS name" {"name": "the person"} {}
    S: SUCCESS {"fields": ["name"]}
    C: PULL {"n": 1000}
    S: SUCCESS {"type": "r"}
    C: COMMIT
    S: SUCCESS {"bookmark": "bm1"}
----
    C: BEGIN {"bookmarks": ["bm1"], "[db]": "*"}
    S: SUCCESS {}
    C: RUN "MATCH (p:Person {name:$name}) RETURN p.name AS name" {"name": "the person"} {}
    S: SUCCESS {"fields": ["name"]}
    C: PULL {"n": 1000}
    S: RECORD ["the person"]
       SUCCESS {"type": "r"}
    C: COMMIT
    S: SUCCESS {"bookmark": "bm2"}
}}
*: RESET

?: GOODBYE
```

Note that I removed the concurrent. We don't expect the driver to open multiple connections to the server, do we?

tests/stub/driver_execute_query/scripts/transaction_chaining_custom_bmm.script
```diff
@@ -0,0 +1,42 @@
...
+            S: SUCCESS {"type": "r"}
+        }}
+        C: COMMIT
+        S: SUCCESS {"bookmark": "bm1"}
+    ----
+        {{
+            C: BEGIN {"bookmarks": ["bm1", "other_db:bm1"], "[db]": "*"}
+        ----
+            C: BEGIN {"bookmarks": ["other_db:bm1", "bm1"], "[db]": "*"}
+        }}
```
robsdedude:
```suggestion
        C: BEGIN {"bookmarks{}": ["bm1", "other_db:bm1"], "[db]": "*"}
```

bigmontz: I don't want to force the order in the bookmark list.

bigmontz: For example, js sends `"other_db:bm1", "bm1"`

robsdedude: I know. That's what the `{}` after the key name is there for. See [Stub script specs (internal doc)](https://docs.google.com/document/d/1QMIVtOTTuNEEZFAVumWGM-BWhLY76wJ56rIAQAisXtA/edit?pli=1#heading=h.horfqne0jrha).

robsdedude (approved):
Fine with me as is :rocket: 

One of my points didn't get discussed
https://github.com/neo4j-drivers/testkit/pull/531#discussion_r997064191
> But why do we even omit them? `SessionRun` e.g., always transmits `params`.

But I can make it work either way in the Python backend.

tests/stub/driver_execute_query/scripts/transaction_chaining_custom_bmm.script
```diff
@@ -0,0 +1,38 @@
...
+C: BEGIN {"bookmarks": ["other_db:bm1"], "[db]": "*"}
+S: SUCCESS {}
+C: RUN "CREATE (p:Person {name:$name}) RETURN p.name AS name" {"name": "a person"} {}
+S: SUCCESS {"fields": ["name"]}
+C: PULL {"n": 1000}
+S: RECORD ["a person"]
+   SUCCESS {"type": "w"}
+C: COMMIT
+S: SUCCESS {"bookmark": "bm1"}
+
```
injectives, line 16:
```suggestion
*: RESET

```

Some drivers might return connection to the pool with `RESET`.

thelonelyvulpes (approved):
Working great on dotnet

PR #532, opened 2022-10-20
Playing around with teamcity output

Prychantovskyi:


robsdedude:
Looks like we didn't get far. We'll pick up the task freshly at a later point in time.

PR #534, opened 2022-11-21
Update Bookmark Manager interface

bigmontz:
Starting with 5.0, Fabric is enabled by default.
It
 1. allows multiple databases to be interacted with in a single transaction, and
 2. ensures bookmarks encode the state of all databases. Hence, the Bookmark Manager can be simplified while also improving its performance (by having to send less bookmarks to the server).

This PR implements the changes need on the bookmark manager API and behaviour.


thelonelyvulpes (approved):
Working well with .NET https://github.com/neo4j/neo4j-dotnet-driver/pull/659

PR #536, opened 2022-11-23
Make more error codes fail fast during the rediscovery

bigmontz:
(Most) drivers will try really hard to fetch a valid RT. This entails going to the next know router if the previous one failed. In this process many errors that are considered unrecoverable in the transaction context are considered retryable. However, a hand full of errors should not be retried to improve user experience. Namely those, that are caused by malformed user-input. Instead of asking all routers if they can make sense of it and ultimately failing with a generic routing update error, it's better to fail fast and give the raw error of the server to the user.

Currently, these errors are
  - `Neo.ClientError.Database.DatabaseNotFound`
 - `Neo.ClientError.Transaction.InvalidBookmark`
 - `Neo.ClientError.Transaction.InvalidBookmarkMixture`

Added three more:
 - `Neo.ClientError.Statement.ArgumentError`: this happens on failed attempts of impersonation (user or permissions missing)
 - `Neo.ClientError.Request.Invalid`: this happens when trying things like impersonating an integer
 - `Neo.ClientError.Statement.TypeError`: this happens when trying things like set a database as an integer in bolt implementation which doesn't support `ROUTE`

### DIFF
--- a/tests/stub/routing/test_routing_v5x0.py
+++ b/tests/stub/routing/test_routing_v5x0.py
@@ -2306,9 +2306,7 @@ class RoutingV5x0(RoutingBase):
             exc.exception.code
         )
 
-    def test_should_fail_with_invalid_route_statement_argument(
-        self
-    ):
+    def test_should_fail_with_invalid_route_statement_argument(self):
         exc = self._test_fast_fail_discover(
             "router_yielding_argument_error.script",
         )
@@ -2318,9 +2316,7 @@ class RoutingV5x0(RoutingBase):
             exc.exception.code
         )
 
-    def test_should_fail_with_invalid_routing_request(
-        self
-    ):
+    def test_should_fail_with_invalid_routing_request(self):
         exc = self._test_fast_fail_discover(
             "router_yielding_invalid_request.script",
         )
@@ -2330,9 +2326,7 @@ class RoutingV5x0(RoutingBase):
             exc.exception.code
         )
 
-    def test_should_fail_with_type_error_request(
-        self
-    ):
+    def test_should_fail_with_type_error_request(self):
         exc = self._test_fast_fail_discover(
             "router_yielding_type_error.script",
         )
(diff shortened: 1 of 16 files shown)


tests/stub/routing/test_routing_v5x0.py
```diff
@@ -2306,6 +2306,42 @@ def test_should_fail_with_routing_failure_on_any_security_discovery_failure(  #
             exc.exception.code
         )
 
+    def test_should_fail_with_invalid_route_statement_argument(
+        self
+    ):
```
robsdedude:
```suggestion
    def test_should_fail_with_invalid_route_statement_argument(self):
```

tests/stub/routing/test_routing_v5x0.py
```diff
@@ -2306,6 +2306,42 @@ def test_should_fail_with_routing_failure_on_any_security_discovery_failure(  #
...
+        )
+
+        self.assertEqual(
+            "Neo.ClientError.Statement.ArgumentError",
+            exc.exception.code
+        )
+
+    def test_should_fail_with_invalid_routing_request(
+        self
+    ):
```
robsdedude:
```suggestion
    def test_should_fail_with_invalid_routing_request(self):
```

tests/stub/routing/test_routing_v5x0.py
```diff
@@ -2306,6 +2306,42 @@ def test_should_fail_with_routing_failure_on_any_security_discovery_failure(  #
...
+        )
+
+        self.assertEqual(
+            "Neo.ClientError.Request.Invalid",
+            exc.exception.code
+        )
+
+    def test_should_fail_with_type_error_request(
+        self
+    ):
```
robsdedude:
```suggestion
    def test_should_fail_with_type_error_request(self):
```

robsdedude (approved):
:zap:

PR #538, opened 2022-12-06
8th and final round of migrating integration tests from Python

robsdedude:
https://github.com/neo4j/neo4j-python-driver/pull/873

### DIFF
--- a/nutkit/frontend/session.py
+++ b/nutkit/frontend/session.py
@@ -80,14 +80,14 @@ class Session:
                     "Should be RetryableTry or RetryableDone but was: %s" % res
                 )
 
-    def read_transaction(self, fn, tx_meta=None, hooks=None, **kwargs):
+    def execute_read(self, fn, tx_meta=None, hooks=None, **kwargs):
         # Send request to enter transactional read function
         req = protocol.SessionReadTransaction(
             self._session.id, txMeta=tx_meta, **kwargs
         )
         return self.process_transaction(req, fn, hooks=hooks)
 
-    def write_transaction(self, fn, tx_meta=None, hooks=None, **kwargs):
+    def execute_write(self, fn, tx_meta=None, hooks=None, **kwargs):
         # Send request to enter transactional read function
         req = protocol.SessionWriteTransaction(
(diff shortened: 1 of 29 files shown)


tests/stub/session_run_parameters/test_session_run_parameters.py
```diff
@@ -116,7 +116,9 @@ def test_tx_meta(self):
             with self.subTest(routing=routing):
                 self._run("tx_meta", routing,
                           session_args=("w",),
-                          run_kwargs={"tx_meta": {"akey": "aval"}})
+                          run_kwargs={
+                              "tx_meta": {"akey": types.CypherString("aval")}
```
injectives, line 120: If we are changing to supplying values as explicit Cypher values, we might want to update it in other places as well. For instance, Java backend does not expect this.

An example of other place: https://github.com/neo4j-drivers/testkit/blob/5.0/tests/stub/driver_parameters/test_bookmark_manager.py#L120

robsdedude: Absolutely! I missed that one, thanks for spotting.

nutkit/frontend/session.py
```diff
@@ -16,7 +16,7 @@ def close(self, hooks=None):
         if not isinstance(res, protocol.Session):
             raise Exception("Should be session but was: %s" % res)
 
-    def run(self, cypher, params=None, tx_meta=None, hooks=None, **kwargs):
+    def run(self, cypher, *, params=None, tx_meta=None, hooks=None, **kwargs):
```
bigmontz, line 19: Does this `*` mean `extra positional params i don't care`?

bigmontz (approved):
❗

nutkit/frontend/session.py
robsdedude: Almost. `*args` (or any other name after `*`) means `collect extra positional arguments in that variable`. If you just write `*` it means from here on, only keyword arguments. FYI: With Python 3.8 they also introduced `/` wich means up until here, positional arguments only ([the PEP](https://peps.python.org/pep-0570/) explaining it).

PR #540, opened 2023-01-23
Fix bookmark manager related test for executeQuery

robsdedude:
Since the test was conceived, the bookmark manager was altered to not track bookmarks per database.


AndyHeap-NeoTech (approved):
All good

PR #542, opened 2023-03-10
Introduce Bolt Protocol 5.1

bigmontz:
Implements the changes need for supporting Bolt v5.1 without the support for user switching and re-auth.

The goal of this changes is verify the support for the new bolt  protocol version in the drivers and the `LOGON` message implementation.

The `LOGOFF` message is not exercised since there isn't any use case for it.

Co-Authored-By: Rouven Bauer <rouven.bauer@neo4j.com>

### DIFF
--- a/tests/stub/versions/scripts/v5x1_return_1.script
+++ b/tests/stub/versions/scripts/v5x1_return_1.script
@@ -1,0 +1,18 @@
+!: BOLT 5.1
+
+C: HELLO {"{}": "*"}
+S: SUCCESS {"server": "#SERVER_AGENT#", "connection_id": "bolt-123456789"}
+A: LOGON {"{}": "*"}
+*: RESET
+{?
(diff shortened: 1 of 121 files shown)


tests/stub/versions/scripts/v5x1_return_1.script
```diff
@@ -0,0 +1,18 @@
+!: BOLT 5.1
```
fbiville, line 1: why not `BOLT #VERSION#` here?

fbiville (approved):

tests/stub/versions/scripts/v5x1_return_1.script
bigmontz: The siblings scripts doesn't expect the version be send as parameters and neither original code for this test, since this versions script are version specific.

PR #543, opened 2023-03-10
Fix issues with linter

bigmontz:
The warning should be log with stacktrace equals to 2, which doesn't make sense in the context. So, the lint for rule B028 was muted.

The except block in wait_for_port was triggering B030 since the construction was not quite right. Changing it to `except (OSError, ConnectionRefusedError):` caused a redundancy since `ConnectionRefusedError` extends `OSError`, so the except block could be reduce to treat only the base class.


robsdedude (approved):
:soap:

PR #546, opened 2023-03-29
Test bolt handshake timeout

robsdedude:
Depends on:
 * https://github.com/neo4j/neo4j-java-driver/pull/1398
 * https://github.com/neo4j/neo4j-python-driver/pull/905

### DIFF
--- a/tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
+++ b/tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
@@ -1,3 +1,5 @@
+from time import perf_counter
+
 from nutkit.frontend import Driver
 import nutkit.protocol as types
 from tests.shared import (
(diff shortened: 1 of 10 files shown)


tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
```diff
@@ -94,6 +100,45 @@ def test_should_work_when_every_step_is_done_in_time(self):
...
+                              connection_acquisition_timeout_ms=1500,
+                              connection_timeout_ms=5000)
+
+        self._session = self._driver.session("r")
+
+        t0 = perf_counter()
+        with self.assertRaises(types.DriverError):
+            list(self._session.run("RETURN 1 AS n"))
+        took = perf_counter() - t0
+        self.assertLess(took, 3)  # give it 1.5 seconds margin
```
thelonelyvulpes, line 121: I really don't like these tests, they are very prone to flakiness.
I see why timing is important here though, we should later consider a way to time it from the testkit backends, as that will be slightly more reliably timed.

thelonelyvulpes (approved):
lgtm.

tests/stub/driver_parameters/test_connection_acquisition_timeout_ms.py
robsdedude: Agreed. Timeouts are generally tough to test. It tends to be slow and flakes out easily :/

PR #551, opened 2023-04-12
Fix integration tests with nightly dev servers

robsdedude:
In particular the agent string check would break with every version bump in the dev build.


fbiville (approved):

fbiville:
Depends on https://github.com/neo4j-drivers/testkit/pull/552

PR #553, opened 2023-04-12
Improve test output in TeamCity

robsdedude:
Multiple improvements:

The first one is that `unittest.TextTestResult` by default logs to stderr (or generally to what ever stream the test runner tells it to), but our custom `TestResult` only ever used `print` with no `file` parameter. So it was always going to stdout. And apparently TC is too dumb to join stdout and stderr without stuff getting all messed up. Or maybe it was because the streams weren't flushed properly. Either way, that looks better now.

Secondly, I changed the custom `TestResult` to print errors in place.

Finally, before, the integration test suites were just called `5.3` etc. But TestKit potentially runs multiple suites with the same server version (but cluster vs single instance, bolt vs neo4j scheme, EE vs CE, ...). So now the suites name contains all that information as well.

### DIFF
--- a/main.py
+++ b/main.py
@@ -24,11 +24,7 @@ import driver
 import neo4j
 import runner
 import settings
-from tests.testenv import (
-    begin_test_suite,
-    end_test_suite,
-    in_teamcity,
-)
+from tests.testenv import in_teamcity
 
 # TODO: Move to docker.py
 networks = ["testkit_1", "testkit_2"]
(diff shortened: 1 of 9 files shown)


fbiville (approved):

AndyHeap-NeoTech (approved):
I do like a good quality of life improvement PR. Nice work.

thelonelyvulpes (approved):

PR #570, opened 2023-05-25
Accept nulls for routing context in Auth tests

thelonelyvulpes:
.NET hasn't implemented the optimisation for not sending null routing contexts.

### DIFF
--- a/tests/stub/authorization/verify_authentication/scripts/v5x0/router.script
+++ b/tests/stub/authorization/verify_authentication/scripts/v5x0/router.script
@@ -1,7 +1,7 @@
 !: BOLT #VERSION#
 !: ALLOW RESTART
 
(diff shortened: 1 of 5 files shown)


tests/stub/authorization/verify_authentication/scripts/v5x0/router.script
```diff
@@ -1,7 +1,7 @@
 !: BOLT #VERSION#
 !: ALLOW RESTART
 
-A: HELLO {"user_agent": "*", "[routing]": {"{}": "*"}, "scheme": "basic", "principal": {"U": "*"}, "credentials": {"U": "*"}, "[realm]": ""}
```
robsdedude: Is this really needed? The router shouldn't be used in the test if there's no routing going on. And if there is, I'd expect the routing context to be not `null`.

(same for the other router script)

robsdedude (approved):
:tulip:

thelonelyvulpes:
Not needed change, mistake on my part.

PR #573, opened 2023-06-15
Add agent parsing from real agent string instead of env 

thelonelyvulpes:
Remove need for integration suites to maintain which version of neo4j they are targeting, Currently we set pipelines to set this value which can make the pipelines fragile i.e. when the pipeline tests 5.10 but the pipeline's env var is set to 5.7.
This PR uses the `get_server_info` API on the 5.X drivers to read the server's agent which should be up to date and thus can be used for us to work out which integration suite we should use.


AndyHeap-NeoTech (approved):
Makes sense to me.

bigmontz (approved):
🍇

robsdedude:
I'm strongly against circular dependencies. Don't use X in the test framework for X (the driver in this case). It'll cause pain sooner or later.

robsdedude:
Also note that aura used to alter the server agent string. So I wouldn't trust it either way.

thelonelyvulpes:
I am not a huge fan of it but we need something more robust than relying people to correctly manage an environment variable.
@robsdedude what could we do instead?

PR #574, opened 2023-06-28
Add tx termination tests

injectives:
These tests verify transaction and its results handling when the transaction is terminated by a failure.


tests/stub/tx_run/test_tx_run.py
```diff
@@ -258,3 +258,28 @@ def test_failed_tx_run_allows_rollback(self):
...
+    def test_should_terminate_tx_after_failed_run(self):
+        self._create_direct_driver()
+        self._server1.start(
+            path=self.script_path("tx_successful_and_failing_streams.script")
+        )
+        self._session = self._driver.session("r", fetch_size=2)
+        tx = self._session.begin_transaction()
+        res = tx.run("RETURN 1 AS n")
+        with self.assertRaises(types.DriverError) as exc:
+            tx.run("invalid")
```
bigmontz, line 316: javascript driver is lazy, so you have to consume the value to get the error
```suggestion
            list(tx.run("invalid"))
```

injectives: Sure. Out of interest, does it pass the test after that?

tests/stub/tx_run/test_tx_run.py
```diff
@@ -258,3 +258,28 @@ def test_failed_tx_run_allows_rollback(self):
...
+        self.assertEqual(exc.exception.code,
+                         "Neo.ClientError.Statement.SyntaxError")
+
+        # while already buffered records may be accessible, there must be no
+        # further PULLs and an exception must be raised
+        with self.assertRaises(types.DriverError):
+            for _i in range(0, 3):
+                res.next()
```
bigmontz: Javascript start to consume the buffer data at this moment , so it tries to get another `PULL` and the script fails since it expects a `RESET`.

However, the result should be closed along with the failed transaction.

tests/stub/tx_run/scripts/tx_error_on_pull.script
```diff
@@ -0,0 +1,19 @@
...
+S: SUCCESS {"fields": ["n"], "qid": 1}
+{{
+    C: PULL {"n": {"Z": "*"}, "[qid]": -1}
+----
+    C: PULL {"n": {"Z": "*"}, "qid": 1}
+}}
+S: RECORD [1]
+   RECORD [2]
+   FAILURE {"code": "Neo.ClientError.MadeUp.Code", "message": "message"}
+*: RESET
```
robsdedude:
```suggestion
+: RESET
```
should we expect the driver to at least `RESET` once?

Same for the other 2 new scripts (won't comment on all).

tests/stub/tx_run/scripts/tx_res0_success_res1_error_on_pull.script
```diff
@@ -0,0 +1,29 @@
...
+S: SUCCESS {"fields": ["n"], "qid": 1}
+{?
+    {{
+        C: PULL {"n": {"Z": "*"}, "[qid]": -1}
+    ----
+        C: PULL {"n": {"Z": "*"}, "qid": 1}
+    }}
+    S: RECORD [1]
+       RECORD [2]
+       SUCCESS {"has_more": true, "type": "r"}
```
robsdedude:
```suggestion
       SUCCESS {"has_more": true}
```
I'm not 100% sure on this one but I don't think the server would transmit the query type if this isn't the summary message without `has_more`. 7687.org does document it this way at least.

tests/stub/tx_run/scripts/tx_error_on_pull.script
robsdedude, line 1: I don't see `tx_error_on_run`. Does such a test already exists? Does it make sense to have?

tests/stub/tx_run/test_tx_run.py
```diff
@@ -258,3 +258,176 @@ def test_failed_tx_run_allows_rollback(self):
...
+            # the streaming result surfaces the termination exception
+            self.assertEqual(exc.exception.code,
+                             "Neo.ClientError.Statement.SyntaxError")
+            self._assert_is_client_exception(exc)
+
+            tx.close()
+            self._session.close()
+            self._session = None
+            self._server1.done()
+        for iterate in ["true", "false"]:
```
robsdedude, line 301:
```suggestion

        for iterate in ["true", "false"]:
```

tests/stub/tx_run/test_tx_run.py
```diff
@@ -258,3 +258,176 @@ def test_failed_tx_run_allows_rollback(self):
...
+            self._server1.start(path=self.script_path(script))
+            self._session = self._driver.session("r", fetch_size=2)
+            tx = self._session.begin_transaction()
+            res = tx.run("RETURN 1 AS n")
+
+            # initiate another stream that fails on RUN
+            with self.assertRaises(types.DriverError) as exc:
+                failed_res = tx.run("invalid")
+                if get_driver_name() in ["javascript"]:
+                    failed_res.next()
```
robsdedude: Can we have a comment why JS needs this extra treatment? And will is remain like this? If it won't remain, this could be a TODO comment

tests/stub/tx_run/test_tx_run.py
```diff
@@ -258,3 +258,176 @@ def test_failed_tx_run_allows_rollback(self):
...
+            with self.assertRaises(types.DriverError) as exc:
+                if iterate == "true":
+                    for _i in range(0, 3):
+                        res.next()
+                else:
+                    fetch_all = types.Feature.OPT_RESULT_LIST_FETCH_ALL
+                    if self.driver_supports_features(fetch_all):
+                        res.list()
+                    else:
+                        # only explicit iteration is tested if the fetch all is
```
robsdedude:
```suggestion
                        # only explicit iteration is tested if fetch all is
```

tests/stub/configuration_hints/test_connection_recv_timeout_seconds.py
```diff
@@ -68,7 +68,7 @@ def _assert_is_timeout_exception(self, e):
     def _assert_is_client_exception(self, e):
         if get_driver_name() in ["java"]:
             self.assertEqual(
-                "org.neo4j.driver.exceptions.ClientException",
+                "org.neo4j.driver.exceptions.TransactionTerminatedException",
                 e.errorType
             )
         elif get_driver_name() in ["ruby"]:
```
robsdedude: Not sure if that should go into a separate PR, but I just realized that `_assert_is_client_exception` and also `_assert_is_timeout_exception` above both don't fail if no error mapping is present. Most other helper methods that make assertions will fail if there is not error mapping for the driver under test. I think that's a good idea so that driver authors will notice that there's work to be done.

tests/stub/tx_run/test_tx_run.py
```diff
@@ -258,3 +258,176 @@ def test_failed_tx_run_allows_rollback(self):
...
+                tx.run("invalid")
+            # new actions on the transaction result in a tx terminated
+            # exception, a subclass of the client exception
+            self._assert_is_tx_terminated_exception(exc)
+
+            tx.close()
+            self._session.close()
+            self._session = None
+            self._server1.done()
+        for iterate in ["true", "false"]:
```
robsdedude, line 393:
```suggestion

        for iterate in ["true", "false"]:
```

tests/stub/tx_run/test_tx_run.py
```diff
@@ -258,3 +258,176 @@ def test_failed_tx_run_allows_rollback(self):
...
+            # the streaming result surfaces the termination exception
+            self.assertEqual(exc.exception.code,
+                             "Neo.ClientError.MadeUp.Code")
+            self._assert_is_client_exception(exc)
+
+            tx.close()
+            self._session.close()
+            self._session = None
+            self._server1.done()
+        for iterate in ["true", "false"]:
```
robsdedude, line 438:
```suggestion

        for iterate in ["true", "false"]:
```

tests/stub/configuration_hints/test_connection_recv_timeout_seconds.py
robsdedude: Similarly with `_assert_is_client_exception` and `_assert_is_tx_terminated_exception` in `tests/stub/tx_run/test_tx_run.py` which are newly added.

robsdedude (approved):
Looks good :clapper: 

Some small comments, but no blockers among them.

tests/stub/tx_run/scripts/tx_error_on_pull.script
injectives: Good point! Fixed.

tests/stub/tx_run/test_tx_run.py
injectives: I think there is a unification pending on that where the driver must report RUN failure on `run` method. I have removed this as the driver can choose to skip this test if that work has not been completed yet.

tests/stub/configuration_hints/test_connection_recv_timeout_seconds.py
injectives: The comparison logic was copied from some other tests, unfortunately I don't remember which. 🤔  I will fix these.

> Not sure if that should go into a separate PR

The `TransactionTerminatedException` is a new type that is introduced in a driver fix that depends on this PR, so I thought it would make sense to have it here.

tests/stub/tx_run/scripts/tx_error_on_pull.script
injectives: I thought that the `test_should_prevent_pull_after_tx_termination_on_run` and `test_should_prevent_discard_after_tx_termination_on_run` already touched on this, but I have added a new `test_should_prevent_run_after_tx_termination_on_run` test as well.

PR #580, opened 2023-07-10
Fix TeamCity output for sub tests

robsdedude:


teamcity/testresult.py
```diff
@@ -37,55 +37,142 @@ def stopTestRun(self):  # noqa: N802
...
+                key = str(sub_test)
+                if key not in test_reports:
+                    test_reports[key] = Report(sub_test, parent_test=test)
+                test_reports[key].errors.append(error)
+            for sub_test, failure in new_failures:
+                key = str(sub_test)
+                if key not in test_reports:
+                    test_reports[key] = Report(sub_test, parent_test=test)
+                test_reports[key].failures.append(failure)
+            breakpoint()
```
bigmontz: Is it an leftover?

bigmontz (approved):
👔

teamcity/testresult.py
robsdedude: :flushed: Good catch. Thanks!

PR #582, opened 2023-07-26
Allow new connection in test_sequence_of_writing_and_reading_tx

injectives:
This update ensures that the driver is permitted to create a new connection for running the second transaction.


bigmontz (approved):
🍎 🍏

PR #583, opened 2023-07-31
Allow another connection in test_should_pass_bookmark_from_tx_to_tx_using_tx_run

injectives:
Some drivers might pick another connection for the subsequent transaction.


robsdedude (commented):
Generally a good idea. But in particular, this change would allow the driver to submit the second query multiple times (over different connections).

Maybe add an assertion to the test using this the script that `RUN` got sent exactly twice.

injectives:
In general, the purpose of this test seems to be to verify that the bookmarks are passed as expected and it would accomplish that. It did not seem to be necessary to be too strict with it, but I have added the requested check as well.

robsdedude (approved):
:link:

PR #585, opened 2023-08-08
Use incremental connection ID in boltstub

robsdedude:


boltstub/bolt_protocol.py
```diff
@@ -35,6 +37,17 @@
...
+auto_bolt_id = 1
+auto_bolt_id_lock = Lock()
+
+
+def next_auto_bolt_id():
+    global auto_bolt_id
+    with auto_bolt_id_lock:
+        current_id = auto_bolt_id
+        auto_bolt_id += 1
+        return f"bolt-{current_id}"
```
bigmontz: Can't you do this? 

```suggestion
    with auto_bolt_id_lock:
        auto_bolt_id += 1
        return f"bolt-{auto_bolt_id}"
```

bigmontz (approved):
🔢

boltstub/bolt_protocol.py
robsdedude: Guess I could. But then I also need init `auto_bolt_id` with `0` so we don't start with `bolt-2` for the first connection.

PR #588, opened 2023-08-10
Add tests for executeQuery BEGIN pipelining

robsdedude:


bigmontz (approved):
🔗

tests/stub/optimizations/test_optimizations.py
```diff
@@ -55,6 +59,51 @@ def test():
...
+        driver = Driver(self._backend, f"bolt://{self._server.address}", auth,
+                        max_tx_retry_time_ms=1500)
+
+        with self.assertRaises(types.DriverError) as exc:
+            driver.execute_query("RETURN 1 AS n")
+
+        driver_name = get_driver_name()
+        if driver_name in ["python"]:
+            expected_error = "<class 'neo4j.exceptions.ServiceUnavailable'>"
+        else:
```
bigmontz:
```suggestion
        elif driver_name in ["javascript"]:
            expected_error = None
            self.assertEqual("ServiceUnavailable", exc.exception.code)
        else:
```

tests/stub/optimizations/test_optimizations.py
```diff
@@ -55,6 +59,51 @@ def test():
...
+        driver.execute_query("RETURN 1 AS n")
+
+        driver.close()
+        self._server.done()
+
+    @driver_feature(types.Feature.API_DRIVER_EXECUTE_QUERY,
+                    types.Feature.OPT_EXECUTE_QUERY_PIPELINING,
+                    types.Feature.BOLT_5_3)
+    def test_execute_query_does_not_pipeline_commit(self):
+        script = "commit_pipeline.script"
```
thelonelyvulpes: This test seems wrong to me, we are verifying that bolt stub shuts down when sent a message in the wrong order.

thelonelyvulpes: this is a case that is implicitly tested, by the above test

thelonelyvulpes: this is the only reliable means of testing this behaviour(currently) so I withdraw this comment.

thelonelyvulpes (approved):

PR #591, opened 2023-08-23
Add dotnet error mappings

RichardIrons-neo4j:
These mappings were missing, hidden due to a weird teamcity-related thing in PyCharm


robsdedude (approved):
. :goal_net:

PR #593, opened 2023-09-12
StubServer: add milliseconds to log output

robsdedude:
This should make it easier to compare timestamps between driver and stub server logs.

### DIFF
--- a/tests/stub/shared.py
+++ b/tests/stub/shared.py
@@ -325,7 +325,8 @@ class StubServer:
         self._wait_for_silence(silence_period)
         res = []
         for line in self._stdout_lines:
-            # lines start with something like "10:08:33  [#EBE0>#2332]  "
+            # lines start with something like
+            # "2023-01-01 10:08:33,420  [#EBE0>#2332]  "
             # plus some color escape sequences and ends on a newline
             line = re.sub(r"\x1b\[[\d;]+m", "", line[:-1])
             line = re.sub(r"^\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2},\d{3}\s+"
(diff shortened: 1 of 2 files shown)


bigmontz (approved):
⏲️

tests/stub/shared.py
```diff
@@ -328,8 +328,8 @@ def get_requests(self, pattern, silence_period=0.1):
             # lines start with something like "10:08:33  [#EBE0>#2332]  "
```
thelonelyvulpes: Should the comments be updated too?

tests/stub/shared.py
```diff
@@ -328,8 +328,8 @@ def get_requests(self, pattern, silence_period=0.1):
...
             # plus some color escape sequences and ends on a newline
             line = re.sub(r"\x1b\[[\d;]+m", "", line[:-1])
-            line = re.sub(r"^\d{2}:\d{2}:\d{2}\s+\[[0-9A-Fa-f#>]+\]\s+", "",
-                          line)
+            line = re.sub(r"^\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2},\d{3}\s+"
```
thelonelyvulpes: Should this be separated by a full stop from seconds like `yyyy-MM-dd HH:mm:ss.SSS` like in ISO8601

thelonelyvulpes (approved):

tests/stub/shared.py
robsdedude: I first thought the logging API didn't provide enough extensibility to do that. I had a second look and made it work. I also decided to omit the date because we're not running stub tests over days or even just hours. Thanks for raising this.

PR #595, opened 2023-09-27
TestTelemetry: Add API_DRIVER_EXECUTE_QUERY to methods which uses `execute_query`

bigmontz:
Not flagging the tests which uses `driver.execute_query` is making Java driver backend fails for reactive and async. The driver only implements this api for sync.

This can be the case of other driver/backend implementations.


robsdedude (approved):
:tada:

PR #596, opened 2023-10-04
Adjust tests related to transaction termination for Javascript

bigmontz:
This PR force the consumption of the result since Javascript Driver doesn't throw on `run`, but only on try to consume the result.

Also add a type mapping to javascript adding error type equals to `Neo4jError`.

### DIFF
--- a/tests/stub/tx_run/test_tx_run.py
+++ b/tests/stub/tx_run/test_tx_run.py
@@ -272,7 +272,7 @@ class TestTxRun(TestkitTestCase):
 
             # initiate another stream that fails on RUN
             with self.assertRaises(types.DriverError) as exc:
-                tx.run("invalid")
+                tx.run("invalid").consume()
             self.assertEqual(exc.exception.code,
                              "Neo.ClientError.Statement.SyntaxError")
             self._assert_is_client_exception(exc)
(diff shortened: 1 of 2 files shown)


robsdedude (approved):
I'd like an additional review by @injectives as he wrote the tests initially. Just to make sure they're still testing what they were intended to.

tests/stub/tx_run/test_tx_run.py
```diff
@@ -313,7 +313,7 @@ def test_should_prevent_discard_after_tx_termination_on_run(self):
 
         # initiate another stream that fails on RUN
         with self.assertRaises(types.DriverError) as exc:
-            tx.run("invalid")
+            tx.run("invalid").consume()
```
injectives, line 316: I believe we have a pending unification to make drivers handle the `RUN` message on `run` method invocation. Is it worth putting the `consume` in a separate block that can be removed at a later stage?

bigmontz: I can check other places where it is done like this. However, in Javascript is impossible to move errors relate to `RUN` message to the `run` method without a big change in the API. `run` returns `Result`, which implements `Promise<{ records, summary}>`. So, when you `run().then().catch()`, you get records and summary or related with RUN or PULL. 

We can argue that we can change `run` the return `Promise<Result>`, however it won't either. Since, `Promise<Result>` is equal to `Promise<Promise<{ records, summary }>>` and it will be reduced to `Promise<{ records, summary }>` and so it will behave exactly how it's behave currently. 

The only way of solving is making `Result` not implement a promise (but a have method called `asPromise` or something like this). The problem with this is the user experience in tx function will be terrible since you will be not able of doing `const { records } = await session.executeRead(tx => tx.run('my query'))` safely. You will have to remember to call `toPromise`.

bigmontz: PR with the changes: https://github.com/neo4j-drivers/testkit/pull/602

PR #598, opened 2023-10-05
Adjust transaction lifetime tests for Python

robsdedude:
Currently, the tests assert that the driver throws the same error with which the other result failed. I personally don't like it because that error has nothing to do with the result stream being used (but only the other failed stream). So instead I decided to throw some generic `ResultFailedError` in Python with the other error as context information for easier debugging.

However, all of this is fool proofing of the drivers. Users will never encounter this error unless they're abusing the driver or doing really funky stuff with it. So I don't have a strong opinion about what exact error we should throw. The main focus should be on preventing the driver from sending invalid messages to the server and ideally throwing any error that's not absolutely cryptic.

### DIFF
--- a/tests/stub/tx_run/test_tx_run.py
+++ b/tests/stub/tx_run/test_tx_run.py
@@ -1,3 +1,5 @@
+import re
+
 from nutkit import protocol as types
 from nutkit.frontend import Driver
 from tests.shared import (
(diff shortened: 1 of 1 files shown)


tests/stub/tx_run/test_tx_run.py
```diff
@@ -343,7 +341,7 @@ def test_should_prevent_run_after_tx_termination_on_run(self):
         self._assert_is_client_exception(exc)
 
         with self.assertRaises(types.DriverError) as exc:
-            tx.run("invalid")
+            tx.run("RETURN 1 AS n")
```
robsdedude, line 344: Note this change.

I felt it was pointless to send a query here which we know will fail with exactly that Syntax error. The driver should throw even before attempting to send anything.

bigmontz (approved):
🧚🏼

PR #603, opened 2023-10-11
.NET transaction terminated validation

thelonelyvulpes:
Update timeout tests to use the transaction terminated exception.
Similar to: https://github.com/neo4j-drivers/testkit/pull/601


RichardIrons-neo4j (approved):
Easy one

PR #606, opened 2023-11-13
Test concurrent auth expired error handling

robsdedude:


robsdedude:
Only thought I'd found a bug in the Python driver. However, it was working alright. The other drivers also appear to do the right thing. Only JS cannot be tested (hence fails) because of the way the TestKit backend has to wrapped around the way the driver handles result streams.

PR #607, opened 2023-11-22
Test transaction config for execute_query

robsdedude:


nutkit/frontend/driver.py
```diff
@@ -101,13 +101,15 @@ def send_and_receive(self, req, timeout=None, hooks=None, *,
         return self.receive(timeout=timeout, hooks=hooks,
                             allow_resolution=allow_resolution)
 
-    def execute_query(self, cypher, params=None,
-                      routing=None, database=None,
-                      impersonated_user=None, bookmark_manager=...):
+    def execute_query(self, cypher, *, params=None, routing=None,
```
bigmontz, line 104: Params should be a positional argument or tests which uses it as positional should be updated.

bigmontz (changes requested):
Make `params` not positional is causing tests failure with `TypeError: execute_query() takes 2 positional arguments but 3 were given`

nutkit/frontend/driver.py
robsdedude: Of course :facepalm: 

Fixed it!

bigmontz (approved):
🩰

thelonelyvulpes (approved):

PR #610, opened 2023-12-14
Add recv_timeout_seconds test to LivenessCheck

bigmontz:
The liveness check `RESET` message should be take in consideration the connection hint `connection.recv_timeout_seconds` to avoid the the check hands forever in case the server doesn't reply.


robsdedude (approved):
:stopwatch:

thelonelyvulpes (approved):

PR #611, opened 2023-12-21
Test server fails reset

robsdedude:


AndyHeap-NeoTech (approved):
Looks sensible to me, but it has been a long time since I dug into testkit tests like this :)

PR #613, opened 2024-02-01
Fix stub server allowing to skip loop1

robsdedude:
A script like
```
{+
    C: Foo
    S: Bar
+}
C: Baz
S: Done
```
allowed the client to skip `Foo` and send `Baz` right away. That's obviously wrong.


StephenCathcart (approved):
🚢

PR #621, opened 2024-07-22
Add support to Bolt 5.6

bigmontz:
Bolt 5.6 fixes the notification description which were missing from Bolt 5.5. Server will not negotiate Bolt 5.5, so tests for this protocol version are not needed.


StephenCathcart (approved):
Looks good 👍

PR #629, opened 2024-09-24
Add Bolt 5.7: GQL-compliant errors

MaxAake:
For testing against during implementation of GQL errors


tests/stub/errors/test_errors.py
```diff
@@ -0,0 +1,331 @@
...
+        )
+        for as_cause in (False, True):
+            with self.subTest(as_cause=as_cause):
+                diagnostic_record = {
+                    "foo": "bar",
+                    "_baz": 1.2,
+                    "OPERATION": None,
+                    "CURRENT_SCHEMA": {"uh": "oh!"},
+                    "OPERATION_CODE": False,
+                    "_classification": 42,
```
injectives, line 309: Just to make sure. If this is not string, we stringify it for the `getRawClassification` method. Right?

robsdedude: I've made it optional in Python and decided to return `None` if it's either missing from the diag. record or isn't a string.

robsdedude (approved):
:mushroom:

PR #650, opened 2025-01-29
Update test_handshake_manifest_server_responses to work for reactive APIs

injectives:


robsdedude (approved):
:left_right_arrow:

tests/stub/transport/test_handshakes.py
```diff
@@ -53,12 +53,12 @@ def _run(self, handshake, handshake_response):
             vars_["#HANDSHAKE_RESPONSE#"] = handshake_response
         with self._get_session(script_path, vars_=vars_) as session:
             session.run("RETURN 1 AS n")
```
MaxAake:
```suggestion
            session.run("RETURN 1 AS n").consume()
```

Testing changes on JS side requires the result to be consumed, otherwise the server is shut down mid-work.

There might be a better fix. But this makes the tests work.

MaxAake (changes requested):

MaxAake (approved):
Looks good, everything seems to work

PR #651, opened 2025-02-05
Home db cache tests: also test executeQuery and explicit tx

robsdedude:


tests/stub/homedb/test_homedb.py
```diff
@@ -204,6 +206,701 @@ def work(tx):
...
+            )
+            route_tracker.assert_new_route_request(**session_args)
+            # Given: driver knows RT for db1
+            self.assertRegex(
+                self._whoami_query(driver, database="db1"),
+                "admin[12]@db1@reader1"
+            )
+            if home_db == "db1":
+                # already knowing RT for db1
+                route_tracker.asset_no_new_route_request()
```
MaxAake: nit: not introduced in this PR, but might as well be fixed.

"asset" should be fixed to "assert", used a 7 times in this file.

tests/stub/homedb/test_homedb.py
```diff
@@ -332,38 +1029,41 @@ def asset_no_new_route_request(self):
...
         )
+        try:
+            records = list(session.run("WHOAMI"))
+            self.assertEqual(len(records), 1)
+            return records[0].values[0].value
+        finally:
+            session.close()
 
     @driver_feature(types.Feature.IMPERSONATION)
     def test_homedb_cache(self):
```
MaxAake, line 1069: This appears to be a remnant of earlier test work. All other tests are nicely defined in the base class and then run for session.run, executeQuery and beginTransaction. This test instead uses execute_read directly and doesn't really cover any functionality not covered by newer, better, standardized tests.

I believe I am the original author of this test and I think it might as well be removed during this rework of the tests. However, as it hasn't really changed in this PR. I think it's fine to merge this and remove the test separately if that is preferred.

MaxAake (approved):
2 comments, but neither was really introduced in this PR nor do they cause issues, so there's a good argument that they be fixed in separate PRs. I leave it up to the author.

tests/stub/homedb/test_homedb.py
robsdedude: It is actually covering some concurrent session scenario that I believe the other tests don't cover. Whether that's worth keeping :shrug: not sure. I'd definitely refrain from removing it in this PR. Feel free to open another PR or we can have a chat/look together in the coming days.

PR #652, opened 2025-02-11
Extend manifest v1 style handshake tests

robsdedude:


tests/stub/transport/test_handshakes.py
```diff
@@ -83,10 +85,61 @@ def test_handshake_manifest_server_responses(self):
...
             ):
                 self._run(server_response, expected_response)
+
+    def _get_newest_bolt_supported_by_driver(self, skip=0, min_version=(5, 0)):
+        # skip: number of newest versions to skip (return n-th newest)
+        # min_version: minimum bolt version to consider
+        all_bolt_versions = [
+            (f, tuple(map(int, f.value.split(":")[-1].split("."))))
+            for f in types.Feature
+            if re.match(r"BOLT_(\d+_)*(\d+)", f.name)
```
MaxAake:
```suggestion
            if re.match(r"^BOLT_(\d+)_(\d+)$", f.name)
```
I feel like a somewhat less permissive regex expression can be used here. This prevents feature names like "EXPLOSIVE_BOLT_5_7" or "BOLT_5_7_KABOOM" from matching the regex.

tests/stub/transport/test_handshakes.py
```diff
@@ -83,10 +85,61 @@ def test_handshake_manifest_server_responses(self):
...
+        ]
+        all_bolt_versions_ge_5_7 = sorted(
```
MaxAake: I don't understand the name of this variable. It reads like "All bolt versions greater or equal to 5.7", so I assume it is a remnant of an earlier phase of work?

MaxAake (commented):
💬

tests/stub/transport/test_handshakes.py
robsdedude: :sweat_smile: good catch. That's a left over from an earlier design :broom:

tests/stub/transport/test_handshakes.py
robsdedude: Yup. That regex is also used in other places in TestKit. I'll adjust them as well, while I'm at it :bow:

tests/stub/transport/test_handshakes.py
```diff
@@ -83,10 +85,57 @@ def test_handshake_manifest_server_responses(self):
...
+            if re.match(r"^BOLT_\d+_\d+$", f.name)
+        ]
+        filtered_bolt_versions = [
+            (f, v) for f, v in all_bolt_versions if v >= min_version
+        ]
+        filtered_bolt_versions.sort(key=lambda x: x[1], reverse=True)
+        for feature, version in filtered_bolt_versions:
+            if self.driver_supports_features(feature):
+                if skip <= 0:
+                    return version
```
MaxAake: There used to be a check here for the protocol to be a 5.x. Without this check this test will blow up once bolt 6.0 exists. Since the server advertises 5.255 -> 5.0 but will expect 6.0.

MaxAake (commented):
💬

tests/stub/transport/test_handshakes.py
robsdedude: :facepalm:

MaxAake (approved):
Looks good!

PR #656, opened 2025-03-12
Stub server: fix auto response for legacy bolt

robsdedude:
* Fix auto response handler for legacy bolt versions checking for HELLO, whereas this message was called INIT before.
 * Clean-up code

### DIFF
--- a/boltstub/bolt_protocol.py
+++ b/boltstub/bolt_protocol.py
@@ -227,7 +227,7 @@ class Bolt1Protocol(BoltProtocol):
     server_agent = "Neo4j/3.3.0"
 
     def get_auto_response(self, request: TranslatedStructure):
-        if request.name in "INIT":
+        if request.name == "INIT":
             return TranslatedStructure(
                 "SUCCESS", b"\x70", {"server": self.server_agent},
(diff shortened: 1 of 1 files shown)


boltstub/bolt_protocol.py
```diff
@@ -227,7 +227,7 @@ class Bolt1Protocol(BoltProtocol):
     server_agent = "Neo4j/3.3.0"
 
     def get_auto_response(self, request: TranslatedStructure):
-        if request.name == "HELLO":
+        if request.name in "INIT":
```
MaxAake:
```suggestion
        if request.name == "INIT":
```
using `in` will make a theoretical request with name `IN` or `INI` pass this check, which I don't think is something we want.

MaxAake (changes requested):
1 minor change request

boltstub/bolt_protocol.py
robsdedude: Whoopsiedoodle. I must have stumbled over my keyboard :smile:

MaxAake (approved):
🚀

PR #658, opened 2025-04-03
Make default db configurable in ITs

robsdedude:


tests/neo4j/test_direct_driver.py
```diff
@@ -193,7 +197,8 @@ def get_names(result_, node=True):
         self._session = self._driver.session("w", database="system",
                                              bookmarks=bookmarks)
         result = self._session.run("SHOW DATABASES")
-        self.assertEqual(get_names(result, node=False), {"system", "neo4j"})
+        self.assertEqual(get_names(result, node=False),
```
MaxAake, line 200: this assertion fails if there are more databases than system and the default. I'm not certain that's a problem, but it feels like something the aura testing usecase might run into.

robsdedude: I kind of expect the ITs to run on a clean DBMS.

robsdedude: Besides the chat we had on Slack, I just realized that this test is also guarded by 
```python
@requires_multi_db_support
@cluster_unsafe_test
```
Both of with alone would make the test be skipped for Aura.

MaxAake (approved):
My question was answered, looks good to me and works well!

PR #661, opened 2025-04-09
Update backend socket timeout

robsdedude:
* Increase the default timeout to 60 seconds for higher resilience at the small risk of longer hanging CI jobs.
 * Make the default timeout configurable via the environment variable `TEST_BACKEND_TIMEOUT`


MaxAake (approved):
Works well!

PR #664, opened 2025-04-25
Fix lints

robsdedude:
With upgrading to Python 3.12 in CI, new lints popped up. While at it, bumping flake8's version as well.


MaxAake (approved):
✅

PR #668, opened 2025-05-05
Fix flaky tx timeout IT

robsdedude:
Test would fail on leader switches and similar because of incorrect error handling.


MaxAake (approved):
✅

PR #669, opened 2025-05-05
[5.0] Fix flaky tx timeout IT

robsdedude:
Test would fail on leader switches and similar because of incorrect error handling.

Backport of https://github.com/neo4j-drivers/testkit/pull/668


MaxAake (approved):
🌸

PR #674, opened 2025-06-04
fix tests for Unification of Connection Acquisition Timeout

MaxAake:
we’ve decided upon unifying the drivers to consider the timeout to count towards everything needed for acquiring a connection. This can include multiple connection acquisitions, ROUTE request, custom resolver/provider calls, etc; anything that’s needed to get a connection for the session.


StephenCathcart (approved):
👍

PR #675, opened 2025-06-05
Remove tests for rejecting some server agent strings

robsdedude:


injectives:
Might be worth adding a test (if its not already there) to fail if no `server` field is available.

StephenCathcart (approved):
🚢

PR #676, opened 2025-06-17
Update test_server_agent

injectives:
Do not force drivers that validate server agent to accept unexpected values.


AndyHeap-NeoTech (approved):
LGTM

PR #683, opened 2025-08-13
Expect Bolt 5.8 instead of 6.0

injectives:


injectives:
https://github.com/neo4j-drivers/testkit/pull/682

PR #688, opened 2025-09-02
Unsupported type stub tests

MaxAake:


nutkit/protocol/cypher.py
```diff
@@ -515,6 +515,62 @@ def __eq__(self, other):
                    for attr in ("dtype", "data"))
 
 
+class CypherUnknownType:
+    r"""
+    A cypher unknown type.
+
+    :param name: The name of the unknown type.
+    :param min_bolt: The minimum required bolt version to transmit the type.
```
injectives: As far as JSON is concerned, I think it should be `minBolt`. Similar to `Node.elementId`.

nutkit/protocol/feature.py
```diff
@@ -90,6 +90,8 @@ class Feature(Enum):
     API_TYPE_TEMPORAL = "Feature:API:Type.Temporal"
     # The driver supports sending and receiving vector data types.
     API_TYPE_VECTOR = "Feature:API:Type.Vector"
+    # The driver supports sending and receiving the unknown data type.
+    API_TYPE_UNKNOWNTYPE = "Feature:API:Type.UnknownType"
```
injectives: `API_TYPE_UNKNOWN_TYPE` maybe?

nutkit/protocol/cypher.py
MaxAake: `min_bolt `is transmitted without any name, not as part of a dictionary. it's written as `min_bolt` in testkit because of python naming conventions, if I tried to call it `minBolt` our pre-commit stylechecks wouldn't let me.

nutkit/protocol/feature.py
MaxAake: yes, absolutely! Fixing that immediately

nutkit/protocol/cypher.py
injectives: I am pretty sure that on the wire we follow camelCase naming though.

MaxAake: Ah, I now see what you mean. Looking deeper it looks like we're inconsistent with the cases of the Cypher types. `CypherRelationship` follows camelCase, whereas `cypherDateTime` follows snake_case 🫤

MaxAake: The new `minimumProtocolMajor` and `minimumProtocolMinor` follow camelCase over the wire!

robsdedude (commented):
Just merged https://github.com/neo4j-drivers/testkit/pull/686/files#diff-6c6229fd9aa35318b990fe02a306cd0ccecf0beec772e9f57711c3e4b886cbc5

Make sure to also update the README changed in the above PR

nutkit/protocol/feature.py
```diff
@@ -90,6 +90,8 @@ class Feature(Enum):
     API_TYPE_TEMPORAL = "Feature:API:Type.Temporal"
     # The driver supports sending and receiving vector data types.
     API_TYPE_VECTOR = "Feature:API:Type.Vector"
+    # The driver supports sending and receiving the unknown data type.
+    API_TYPE_UNKNOWN_TYPE = "Feature:API:Type.UnknownType"
```
robsdedude: Alphabetical order please :)

boltstub/simple_jolt/v3/codec.py
```diff
@@ -93,6 +96,64 @@ def _encode_full(cls, value, encode_cb, human_readable):
...
+                ' second element of array after sigil "UT"'
+            )
+        minimum_protocol_major = value[1]
+        if not isinstance(value[2], int):
+            raise JOLTValueError(
+                "Expecting minimum bolt minor version as"
+                ' third element of array after sigil "UT"'
+            )
+        minimum_protocol_minor = value[2]
+        if not isinstance(value[3], dict):
```
robsdedude: To make stub script more resilient and to make it easier to spot typos, I suggest to sanitize the extra dict even more. Make sure there are no unknown/unexpected entries in `extra`. (same for the rusty stub server)

boltstub/packstream.py
```diff
@@ -416,6 +417,16 @@ def _from_jolt_v3_type(cls, jolt: jolt_v3_types.JoltType):
                 jolt.data,
                 packstream_version=3,
             )
+        if isinstance(jolt, jolt_v3_types.JoltUnknownType):
+            extra = {"message": jolt.message}
```
robsdedude: I don't think the real server would send `extra: {"message": null}`, Instead it would probably send an empty dict. Which is what your Rust implementation does, btw. So at least both should be consistent.

boltstub/simple_jolt/v3/jolt_types.py
```diff
@@ -64,7 +64,28 @@ def __init__(self, dtype, data):
...
+        self.name = name
+        self.minimum_protocol_major = minimum_protocol_major
+        self.minimum_protocol_minor = minimum_protocol_minor
+        self.message = message
+
+
 __all__ = [
     JoltType,
     JoltVector,
+    JoltUnknownType,
```
robsdedude:
```suggestion
    JoltUnknownType,
    JoltVector,
```
:abc:

boltstub_rs/README.md
```diff
@@ -693,6 +693,20 @@ Example:
 {"V": ["i16", "FF 00"]}
 ```
 
+### JOLT UnknownType
+*NOTE*:  
+Standard JOLT does not support Unknown Types.
+
+**Simple**: **not supported**
+
+**Full**: `{"UT": ["<type name>", minimum_protocol_major, minimum_protocol_minor, #, <extra dictionary with optional message>]}`
```
robsdedude: Taking a step back: alternative design ideas:
 * Have the JOLT syntax look like this and pass the extra dict *exactly* as is (not only the message entry).
   * Pro: most flexible
   * Con: allows for non-bolt-compliant messages (also by accident)
 * Have the last entry either be and optional string (optional either meaning that it can be `null` or that it can be omitted making the list only 3 elements long)
   * Pro: more closely resembles bolt's specs while not forcing users to repeat a magic string "message"
   * Con: enforces bolt compliance (there's always `S: <RAW> ...` for hand-crafted nonsense) 

Obviously keeping it as is is also an option. Just food for thought.

boltstub_rs/src/parser.rs
```diff
@@ -990,6 +990,10 @@ fn transcode_jolt_value(
             let bolt_path = JoltVector::parse(value, jolt_version, config)?;
             IsJoltValue::Yes(PackStreamValue::Struct(bolt_path.into_struct()))
         }
+        JoltSigil::UnknownType => {
+            let bolt_path = JoltUnknownType::parse(value, jolt_version, config)?;
+            IsJoltValue::Yes(PackStreamValue::Struct(bolt_path.into_struct()))
```
robsdedude:
```suggestion
            let bolt_unknown_type = JoltUnknownType::parse(value, jolt_version, config)?;
            IsJoltValue::Yes(PackStreamValue::Struct(bolt_unknown_type.into_struct()))
```

boltstub_rs/src/values/bolt_struct/unknown_type.rs
```diff
@@ -0,0 +1,140 @@
...
+
+use serde_json::Value as JsonValue;
+
+use super::_parsing::{check_last_json_field, next_json_field, next_pack_stream_field};
+use crate::bolt_version::JoltVersion;
+use crate::parse_error::ParseError;
+use crate::parser::ActorConfig;
+use crate::values::bolt_struct::TAG_UNKNOWN_TYPE;
+use crate::values::pack_stream_value::{PackStreamStruct, PackStreamValue};
+use indexmap::IndexMap;
```
robsdedude:
```suggestion
use serde_json::Value as JsonValue;
use indexmap::IndexMap;

use super::_parsing::{check_last_json_field, next_json_field, next_pack_stream_field};
use crate::bolt_version::JoltVersion;
use crate::parse_error::ParseError;
use crate::parser::ActorConfig;
use crate::values::bolt_struct::TAG_UNKNOWN_TYPE;
use crate::values::pack_stream_value::{PackStreamStruct, PackStreamValue};
```

groups: 
stdlib
3rd party
1st pary

boltstub_rs/src/values/bolt_struct/unknown_type.rs
```diff
@@ -0,0 +1,140 @@
...
+    pub(crate) fn parse(
+        v: JsonValue,
+        jolt_version: JoltVersion,
+        config: &ActorConfig,
+    ) -> Result<Self, ParseError> {
+        if !matches!(jolt_version, JoltVersion::V3) {
+            return Err(ParseError::new(format!(
+                "Sigil \"UT\" (unknown type) can only be parsed in JoltVersion::V3, using {jolt_version:?}"
+            )));
+        }
```
robsdedude: Either (admittedly ugly)
```suggestion
        match jolt_version {
            JoltVersion::V1 | JoltVersion::V2 => {
                return Err(ParseError::new(format!(
                    "Sigil \"UT\" (unknown type) can only be parsed in JoltVersion::V3, \
                    using {jolt_version:?}"
                )));
            }
            JoltVersion::V3 => {}
        }
```
or 
```suggestion
        if jolt_version < JoltVersion::V3 {
            return Err(ParseError::new(format!(
                "Sigil \"UT\" (unknown type) can only be parsed in JoltVersion::V3, \
                    using {jolt_version:?}"
            )));
        }
```
To make the code more future proof. The former will cause a compiler error when a new Jolt version is introduced, the latter assumes that future versions are unlikely to change the semantics of the unknown type. The latter also requires to derive `PartialOrd` (and while you're at it Ord` wouldn't hurt either) for `JoltVersion`.

boltstub_rs/src/values/bolt_struct/unknown_type.rs
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    pub(crate) fn into_struct(self) -> PackStreamStruct {
+        let mut extra_map = IndexMap::with_capacity(1);
+        if self.message.is_some() {
+            extra_map.insert(
+                String::from("message"),
+                PackStreamValue::String(self.message.unwrap()),
+            );
+        }
```
robsdedude:
```suggestion
        if let Some(message) = self.message {
            extra_map.insert(String::from("message"), PackStreamValue::String(message));
        }
```
100% panic-path-free[^1] code :relieved: 

[^1]: bearing OOMs and other fatal things.

boltstub_rs/src/values/bolt_struct/unknown_type.rs
```diff
@@ -0,0 +1,140 @@
...
+            tag: TAG_UNKNOWN_TYPE,
+            fields,
+        }
+    }
+}
+#[derive(Debug, Clone)]
+pub(super) struct BoltUnknownType(pub(super) JoltUnknownType);
+impl BoltUnknownType {
+    pub(super) fn from_struct(s: &PackStreamStruct, jolt_version: JoltVersion) -> Option<Self> {
+        if !matches!(jolt_version, JoltVersion::V3) || s.tag != TAG_UNKNOWN_TYPE {
```
robsdedude: (as above) please make jolt version matching for future-proof.

boltstub_rs/src/values/bolt_struct/unknown_type.rs
```diff
@@ -0,0 +1,140 @@
...
+            return None;
+        }
+        let mut fields = s.fields.iter();
+        let name: &str = next_pack_stream_field(&mut fields)?;
+        let minimum_protocol_major: i64 = next_pack_stream_field(&mut fields)?;
+        let minimum_protocol_minor: i64 = next_pack_stream_field(&mut fields)?;
+        let extra: &IndexMap<String, PackStreamValue> = next_pack_stream_field(&mut fields)?;
+        let PackStreamValue::String(ref message) = extra["message"] else {
+            return None;
+        };
```
robsdedude: Iirc, this code wouldn't ever be reached because this it the path where the server reads structs. And an unknown type struct isn't supposed to be sent to the server. Anyway, it's been implemented for other out-only types (like Nodes), too. So I think it's good to also implement it here. However, it should probably not reject the struct if the *optional* message is omitted.

boltstub_rs/src/values/bolt_struct/unknown_type.rs
```diff
@@ -0,0 +1,140 @@
...
+            this: &'a JoltUnknownType,
+        }
+
+        impl Display for JoltFormatter<'_> {
+            fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+                f.write_str(r#"{"UT": [""#)?;
+                f.write_str(&self.this.name)?;
+
+                f.write_str(r#"", ""#)?;
+                f.write_str(&self.this.minimum_protocol_major.to_string())?;
```
robsdedude:
```suggestion
                Display::fmt(&self.this.minimum_protocol_major, f)?;
```
I reckon this saves at least on String allocation (not tested).

tests/stub/datatypes/scripts/v6x0/echo_unknown.script
```diff
@@ -0,0 +1,16 @@
+!: BOLT 6.0
+
+A: HELLO {"{}": "*"}
+A: LOGON {"{}": "*"}
+*: RESET
+C: RUN {"U": "RETURN 1 as one"} {"{}": "*"} {"{}": "*"}
```
robsdedude: nit over-9000/10

cypher style guide
```suggestion
C: RUN {"U": "RETURN 1 AS one"} {"{}": "*"} {"{}": "*"}
```
(requires changing test code as well :grimacing: )

tests/stub/datatypes/test_unknown_type.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def test_unknown_type_subtests(self):
+
+        script = "echo_unknown.script"
+        for (name, minimum_protocol_major, minimum_protocol_minor, extra) in (
+            ("encrypted_value", 6, 10, None),
+            ("encrypted_value", 6, 10, {"message": "test message"}),
+            (
+                "encrypted_value", 6, 10,
+                {"message": "test message", "junk data": "junk"}
```
robsdedude: The way the stub server changes are, rn, this test is useless as the sever will silently swallow the extra key (see comment above).

robsdedude (changes requested):
:rocket:

boltstub/simple_jolt/v3/codec.py
robsdedude: Actually, read https://github.com/neo4j-drivers/testkit/pull/688#discussion_r2348919068 first

boltstub_rs/src/values/bolt_struct/unknown_type.rs
robsdedude, line 1: Consider adding some unit tests (also to the Python implementation).

nutkit/protocol/cypher.py
```diff
@@ -515,6 +515,80 @@ def __eq__(self, other):
...
+    :param minimum_protocol_major: The major bolt version needed for the type.
+    :param minimum_protocol_minor: The minor bolt version needed for the type.
+    :param message: An optional message to the user.
+    """
+
+    def __init__(
+        self, name, minimumProtocolMajor, minimumProtocolMinor, message=None
+    ):
+        self.name = str(name)
+        self.minimum_protocol_major = minimumProtocolMajor
```
injectives: While we have decided to have 2 fields on Bolt level, are we still expecting the driver API to expose them together as a string? If yes, we should probably have the same here.

robsdedude: The ADR leaves this open to fit in with the rest of the driver API. As a guiding help consider what type a given driver uses for the protocol version in the summary API.

But generally I agree. Since that might be a string in some drivers, having it be a string in the TestKit protocol makes sense. It's slightly easier for a backend to turn some other type (e.g. 2 numbers) into a string that it is to turn a string into two numbers.

MaxAake: Adjusting to a single "minimumProtocol" string. 👍🏻

nutkit/protocol/feature.py
```diff
@@ -88,6 +88,8 @@ class Feature(Enum):
     API_TYPE_SPATIAL = "Feature:API:Type.Spatial"
     # The driver supports sending and receiving temporal data types.
     API_TYPE_TEMPORAL = "Feature:API:Type.Temporal"
+    # The driver supports sending and receiving the unsupported type data type.
+    API_TYPE_UNSUPPORTED_TYPE = "Feature:API:Type.UnsupportedType"
```
injectives, line 92: Isn't this part of Bolt 6.0? Maybe we don't need a feature flag for it.

injectives: As far as I am aware, supporting this is not optional.

MaxAake: That's a good point, the same is true for Vectors and we made a feature flag for those as well. It's mainly for development purposes, that way someone working on their 6.x branch won't start seeing red tests the moment we merge.

But it's worth considering if we want to use flags like that.

nutkit/protocol/cypher.py
```diff
@@ -515,6 +515,71 @@ def __eq__(self, other):
                    for attr in ("dtype", "data"))
 
 
+class CypherUnsupportedType:
+    r"""
+    A cypher unknown type.
+
+    :param name: The name of the unknown type.
```
robsdedude:
```suggestion
    A cypher unsupported type.

    :param name: The name of the unsupported type.
```

boltstub_rs/src/values/bolt_struct/unsupported_type.rs
```diff
@@ -0,0 +1,258 @@
...
+impl JoltUnsupportedType {
+    pub(crate) fn parse(
+        v: JsonValue,
+        jolt_version: JoltVersion,
+        config: &ActorConfig,
+    ) -> Result<Self, ParseError> {
+        match jolt_version {
+            JoltVersion::V1 | JoltVersion::V2 => {
+                return Err(ParseError::new(format!(
+                    "Sigil \"UT\" (unknown type) can only be parsed in JoltVersion::V3,
```
robsdedude:
```suggestion
                    "Sigil \"UT\" (unsupported type) can only be parsed in JoltVersion::V3,
```

boltstub/tests/test_packstream.py
```diff
@@ -278,6 +278,38 @@
...
     ),
+    # Unsupported Type
+    (
+        1,
+        [Structure(b"\x3F", "Quantum Integer", 6, 10, packstream_version=1)],
+        TypeError
+    ),
+    (
+        2,
+        [Structure(b"\x3F", "Quantum Integer", 6, 10, packstream_version=3)],
```
robsdedude:
```suggestion
        [Structure(b"\x3F", "Quantum Integer", 6, 10, packstream_version=2)],
```

boltstub_rs/README.md
```diff
@@ -729,4 +729,21 @@ Stubscript in contrast considers each change, regardless whether to a Structure
  * Default for Bolt versions: 6.0+
  * Changes:
    * Add `Vector` type
-   * Add `UnknownType` type
+   * Add `UnsupportedType` type
+tructure representations dependent on the negotiated Bolt version.
```
robsdedude: omething is missing and misaligned here.

boltstub_rs/src/values/bolt_struct/unsupported_type.rs
```diff
@@ -0,0 +1,258 @@
...
+        Ok(Self {
+            name,
+            minimum_protocol_major,
+            minimum_protocol_minor,
+            message: Some(message),
+        })
+    }
+
+    pub(crate) fn into_struct(self) -> PackStreamStruct {
+        let mut extra_map = IndexMap::with_capacity(1);
```
robsdedude: Nit level >9000: if the `message` is `None`, this is an unnecessary allocation (assuming empty maps don't allocate). Obviously not a blocker.

boltstub_rs/src/values/bolt_struct/unsupported_type.rs
```diff
@@ -0,0 +1,258 @@
...
+        match jolt_version {
+            JoltVersion::V1 | JoltVersion::V2 => return None,
+            JoltVersion::V3 => {}
+        }
+        let mut fields = s.fields.iter();
+        let name: &str = next_pack_stream_field(&mut fields)?;
+        let minimum_protocol_major: i64 = next_pack_stream_field(&mut fields)?;
+        let minimum_protocol_minor: i64 = next_pack_stream_field(&mut fields)?;
+        let extra: &IndexMap<String, PackStreamValue> = next_pack_stream_field(&mut fields)?;
+        let PackStreamValue::String(ref message) = extra["message"] else {
```
robsdedude: Panics for valid input

```rust
    #[rstest]
    #[case(None)]
    #[case(Some(String::from("From the future")))]
    fn test_struct_parsing(#[case] message: Option<String>) {
        let extra = match message.clone() {
            Some(message) => IndexMap::from([(
                String::from("message"),
                PackStreamValue::String(message.clone()),
            )]),
            None => IndexMap::new(),
        };
        let structure = PackStreamStruct {
            tag: TAG_UNSUPPORTED_TYPE,
            fields: vec![
                PackStreamValue::String(String::from("Quantum Integer")),
                PackStreamValue::Integer(6),
                PackStreamValue::Integer(10),
                PackStreamValue::Dict(extra),
            ],
        };
        assert_eq!(
            BoltUnsupportedType::from_struct(&structure, JoltVersion::V3),
            Some(BoltUnsupportedType(JoltUnsupportedType {
                name: String::from("Quantum Integer"),
                minimum_protocol_major: 6,
                minimum_protocol_minor: 10,
                message,
            }))
        )
    }
```
(requires `#[cfg_attr(test, derive(PartialEq))]` on `BoltUnsupportedType` and `JoltUnsupportedType`)

boltstub_rs/src/values/bolt_struct/unsupported_type.rs
```diff
@@ -0,0 +1,258 @@
...
+                f.write_str(&self.this.name)?;
+
+                f.write_str(r#"", "#)?;
+                Display::fmt(&self.this.minimum_protocol_major, f)?;
+
+                f.write_str(r#", "#)?;
+                Display::fmt(&self.this.minimum_protocol_minor, f)?;
+                if self.this.message.is_some() {
+                    f.write_str(r#", ""#)?;
+                    f.write_str(&self.this.message.clone().expect("TEST"))?;
```
robsdedude: "TEST" is not a great panic message :upside_down_face: might as well simply use `unwrap` then. Even better, however: use `if let Some(...) = self.this.message`

nutkit/protocol/feature.py
injectives: I know. Personally, I would expect the driver to support both vectors and unsupported type so long as it claims support for Bolt 6.0.

tests/stub/datatypes/test_unsupported_type.py
```diff
@@ -53,110 +53,88 @@ def _session(self, driver):
         finally:
             session.close()
 
-    def test_unknown_type_subtests(self):
+    def test_unsupported_type_subtests(self):
```
robsdedude:
```suggestion
    def test_unsupported_type(self):
```
or what is the thinking behind the "subtests" suffix?

MaxAake: A remnant from when I converted these tests from separate but almost identical to subtests, no reason to have it there anymore. thanks!

boltstub_rs/src/values/bolt_struct/unsupported_type.rs
MaxAake: Lesson learned, rust compiler is cool, but it can't always tell you the best way to do something 😂 

Though I can't blame it for setting a panic message to "TEST", that one's on me

boltstub_rs/src/values/bolt_struct/unsupported_type.rs
```diff
@@ -66,9 +67,12 @@ impl JoltUnsupportedType {
...
 
     pub(crate) fn into_struct(self) -> PackStreamStruct {
-        let mut extra_map = IndexMap::with_capacity(1);
+        let mut extra_map;
         if let Some(message) = self.message {
+            extra_map = IndexMap::with_capacity(1);
             extra_map.insert(String::from("message"), PackStreamValue::String(message));
+        } else {
+            extra_map = IndexMap::with_capacity(0);
         }
```
robsdedude:
```suggestion
        let extra_map = match self.message {
            Some(message) => {
                IndexMap::from([(String::from("message"), PackStreamValue::String(message))])
            }
            None => IndexMap::new(),
        };
```
slightly preferable, I find. a) doesn't leave a variable mut for longer than needed b) (in my view) ever so slightly more readable.

boltstub/packstream.py
```diff
@@ -562,6 +569,20 @@ def _to_jolt_v3_type(self):
...
+                    protocol_minimum_minor, extra
+                ) = self.fields
+                return jolt_v3_types.JoltUnsupportedType(
+                    name, protocol_minimum_major,
+                    protocol_minimum_minor, extra["message"]
+                )
+            name, protocol_minimum_major, protocol_minimum_minor = self.fields
+            return jolt_v3_types.JoltUnsupportedType(
+                name, protocol_minimum_major, protocol_minimum_minor
+            )
```
robsdedude: According to the ADR, the struct always has 4 fields, but `extra["message"]` might result in a `KeyError` :upside_down_face:

MaxAake: corrected, along with the faulty test that led to this miss

robsdedude (approved):
:man_singer: 
:guitar:

PR #690, opened 2025-09-11
Fix script incompatibility with lazy connection closing

robsdedude:
Some drivers (e.g., our Go driver) may choose to close connection in a background task. Therefore, there's a race condition if the driver is instructed to shut down and right after, the stub server is ended if the script asserts that the connection reaches its end with `GOODBYE`.


MaxAake (approved):
✅

PR #691, opened 2025-10-02
ITs: add vector tests

robsdedude:


tests/neo4j/datatypes/_base.py
```diff
@@ -30,14 +30,21 @@ def _create_driver_and_session(self):
...
-            result = tx.run("RETURN $x as y", params={"x": val})
+            result = tx.run("RETURN $x AS y", params={"x": val})
             record_ = result.next()
             assert isinstance(result.next(), types.NullRecord)
             return record_
 
         record = self._session.execute_read(work)
         self.assertEqual(record, types.Record(values=[val]))
 
+    def _send_value(self, val):
```
MaxAake: I'm wondering if there is a better name for this function?
```suggestion
    def _attempt_send_value_in_param(self, val):
```
I don't know, it's a little wordy, but `send_value` isn't very descriptive.

MaxAake (commented):
💬 ❓

MaxAake (approved):
✅

PR #692, opened 2025-10-23
Aura ITs: loosen up version assumptions

robsdedude:
* Allow bolt version changes upwards
 * Bolt 6.0 not yet available on aura


MaxAake (approved):
✅

PR #695, opened 2025-10-23
[5.0] Aura ITs: loosen up version assumptions

robsdedude:
* Allow bolt version changes upwards
 * Bolt 6.0 not yet available on aura
 * Fix using strings for version comparisons
 
 Port of https://github.com/neo4j-drivers/testkit/pull/692 plus some extra improvements from the 6.x branch.

PR #697, opened 2025-10-24
[5.0] Fix Aura ITs bug

robsdedude:
Port of https://github.com/neo4j-drivers/testkit/pull/696

Amends https://github.com/neo4j-drivers/testkit/pull/695

Fixes DRIVERS-67


MaxAake (approved):
✅

PR #698, opened 2025-11-14
Rusty stub: Fix cancellation shutdown

robsdedude:
* If a server action inside a loop fails (e.g., because it was canceled), the error should be propagated straight away and not deferred until the whole block has been handled.

* Mark action lines as skippable

  When only action lines (server lines, Python lines, etc.) are left, the block can be considered skipable as no client-side action is required.
  
Closes DRIVERS-213


MaxAake (approved):
✅

PR #702, opened 2026-02-04
Rust stub: bump dependencies

robsdedude:


AndyHeap-NeoTech (approved):
Thats a lot of version bumps :) Can't see any errors with it.

PR #703, opened 2026-02-05
Rust stub: crank clippy linter to pedantic

robsdedude:
Closes: DRIVERS-256

### DIFF
--- a/boltstub_rs/src/values/bolt_struct/date.rs
+++ b/boltstub_rs/src/values/bolt_struct/date.rs
@@ -33,26 +33,24 @@ impl JoltDate {
                 .as_str(),
         )
         .expect("regex enforces year to be i32");
-        let month = captures
-            .get(2)
-            .map(|m| u32::from_str(m.as_str()).expect("regex enforces month to be u32"))
-            .unwrap_or(1);
-        let day = captures
-            .get(3)
-            .map(|d| u32::from_str(d.as_str()).expect("regex enforces day to be u32"))
-            .unwrap_or(1);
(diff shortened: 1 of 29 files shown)


boltstub_rs/src/values/bolt_struct/date.rs
```diff
@@ -33,26 +33,24 @@ impl JoltDate {
...
         .expect("regex enforces year to be i32");
-        let month = captures
-            .get(2)
-            .map(|m| u32::from_str(m.as_str()).expect("regex enforces month to be u32"))
-            .unwrap_or(1);
-        let day = captures
-            .get(3)
-            .map(|d| u32::from_str(d.as_str()).expect("regex enforces day to be u32"))
-            .unwrap_or(1);
+        let month = captures.get(2).map_or(1, |m| {
```
AndyHeap-NeoTech, line 36: Did the linter suggest this change. If so surprises me as usually multiline style is the things go.

AndyHeap-NeoTech (approved):

boltstub_rs/src/values/bolt_struct/date.rs
robsdedude: It did. The linter was pushing for using `map_or(a, b)` instead of `.map(b).unwrap_or(a)`.
The way this wrapped across multiple lines in this case was more of a side-effect than the linter prefering fewer lines per se. The formatter appears to have a heuristic considering `captures.get(2).map_or(` short enough to put it on one line 🤷

PR #705, opened 2026-03-06
Improve profile and plan stub tests

robsdedude:
* Allow missing children key, null children value, empty list children value are now considered equivalent.
  * Add tests for bolt 6.0 where the structure of profiles and plans has changed a fair bit. Most prominently will `children` be omitted by the bolt server when it's empty.
  * Adding tests under new feature flag `"Feature:API:Summary:Profile:OptionalStats"`:  
    Drivers should be able to differentiate missing profile stats (e.g., `dbHits = 0` has a different meaning that absent `dbHits`)
  
  Closes: DRIVERS-281
  
  Depends on:
   * https://github.com/neo4j/neo4j-javascript-driver/pull/1397
   * https://github.com/neo4j/neo4j-java-driver/pull/1731

### DIFF
--- a/tests/stub/summary/test_summary.py
+++ b/tests/stub/summary/test_summary.py
@@ -1433,7 +1433,7 @@ class TestSummaryPlan4x4(_TestSummaryBase):
             "summary_with_plan.script",
             vars_={"#PLAN#": json.dumps(plan)},
         )
-        self.assert_plan_equal(plan, summary.plan)
+        self.assert_plan_equal(summary.plan, plan)
 
     def test_profile(self):
         profile = {
@@ -1556,7 +1556,7 @@ class TestSummaryPlan6x0(_TestSummaryBase):
             "summary_with_plan.script",
             vars_={"#PLAN#": json.dumps(plan)},
         )
-        self.assert_plan_equal(plan, summary.plan)
+        self.assert_plan_equal(summary.plan, plan)
 
     def test_profile(self):
         profile = {
--- a/tests/stub/summary/test_summary.py
+++ b/tests/stub/summary/test_summary.py
@@ -49,6 +49,31 @@ class _TestSummaryBase(TestkitTestCase):
             list(result)
             return result.consume()
 
+    def assert_plan_equal(self, actual, expected):
+        def adjust_expected(actual_child, expected_child):
+            # drivers are free to represent lack of children with either:
+            #   * an empty list
+            #   * a null value
+            #   * lack of the key
+            expected_children = expected_child.get("children")
+            expected_has_children = expected_children not in (None, [])
+            actual_children = actual_child.get("children")
+            actual_has_children = actual_children not in (None, [])
+
+            if not expected_has_children and not actual_has_children:
+                expected_child.pop("children", None)
+                actual_child.pop("children", None)
+                return
+
+            expected_children = expected_children or []
+            actual_children = actual_children or []
+            if len(expected_children) == len(actual_children):
+                for ac, ec in zip(actual_children, expected_children):
+                    adjust_expected(ac, ec)
+
+        adjust_expected(actual, expected)
+        self.assertEqual(actual, expected)
+
 
 class _TestSummaryDiscardMixin(_TestSummaryBase):
     def _get_summary(self, script, vars_=None):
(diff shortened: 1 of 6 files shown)


tests/stub/summary/test_summary.py
```diff
@@ -49,6 +49,46 @@ def _get_summary(self, script, vars_=None):
...
             return result.consume()
 
+    def assert_plan_equal(self, actual, expected):
+        missing_stats_detectable = self.driver_supports_features(
+            types.Feature.API_SUMMARY_PROFILE_OPTIONAL_STATS
+        )
+
+        def adjust_expected(actual_child, expected_child):
+            for key in [k for k, v in expected_child.items() if v is None]:
+                if key not in expected_child:
```
AndyHeap-NeoTech: Is this correct? You are iterating over expected_child looking for values that are none. You then check that the values key is _not_ in expected child. I think this might be backwards making the loop do nothing?

tests/stub/summary/test_summary.py
```diff
@@ -49,6 +49,46 @@ def _get_summary(self, script, vars_=None):
...
+                    expected_child.pop(key)
+            for key in (k for k, v in expected_child.items() if v is None):
```
AndyHeap-NeoTech: Isn't this also effectively an no-op loop. If a value is none, you then set it to none again.

tests/stub/summary/test_summary.py
```diff
@@ -1368,28 +1408,29 @@ def test_plan(self):
...
-            "identifiers": ["n"]
+            "identifiers": ["n"],
         }
         summary = self._get_summary(
             "summary_with_plan.script",
-            vars_={"#PLAN#": json.dumps(plan)}
+            vars_={"#PLAN#": json.dumps(plan)},
         )
-        self.assertEqual(summary.plan, plan)
+        self.assert_plan_equal(plan, summary.plan)
```
AndyHeap-NeoTech: assert_plan_equal takes parameter order of (actual, expected). This might be backwards, isn't plan the one you are expecting and summary.plan the one coming from the driver?

Worth checking this ordering in all locations.

tests/stub/summary/test_summary.py
```diff
@@ -1428,22 +1469,331 @@ def test_profile(self):
...
             "identifiers": ["n"],
             "operatorType": "ProduceResults@neo4j",
-            "rows": 1
+            "rows": 1,
+        }
+        summary = self._get_summary(
+            "summary_with_profile.script",
+            vars_={"#PROFILE#": json.dumps(profile)},
+        )
+        self.assert_plan_equal(summary.profile, profile)
```
AndyHeap-NeoTech, line 1489: Ordering seems correct here.

tests/stub/summary/test_summary.py
```diff
@@ -49,6 +49,46 @@ def _get_summary(self, script, vars_=None):
             list(result)
             return result.consume()
 
+    def assert_plan_equal(self, actual, expected):
```
AndyHeap-NeoTech, line 53: An observation. The method is mutating the dictionaries being passed in. I'm concerned that this could lead to flaky tests or knock on impacts of one test on another. Especially when combined with the parameter ordering differences on calls to the method that I note below.

robsdedude: I don't have a strong opinion on this. The changes are kind of idempotent. But I'll put a deep copy in there for peace of mind. It's just test code and not exactly performance critical.

tests/stub/summary/test_summary.py
```diff
@@ -49,6 +50,48 @@ def _get_summary(self, script, vars_=None):
...
+            #   * a null value
+            #   * lack of the key
+            expected_children = expected_child.get("children")
+            expected_has_children = expected_children not in (None, [])
+            actual_children = actual_child.get("children")
+            actual_has_children = actual_children not in (None, [])
+
+            if not expected_has_children and not actual_has_children:
+                expected_child.pop("children", None)
+                actual_child.pop("children", None)
```
AndyHeap-NeoTech: Minor point here. If we are adjusting both expected and actual then the method name is misleading.

AndyHeap-NeoTech (approved):
Looks good, just a minor observation after the last commit.

tests/stub/summary/test_summary.py
robsdedude: Here's my suggested fix. Does this look correct to you? It's slightly less elegant which is why I didn't go for it initially.

```python
expected_children = expected_child.get("children")
expected_has_no_children = expected_children in (None, [])
actual_children = actual_child.get("children")
actual_has_no_children = actual_children in (None, [])

if expected_has_no_children and actual_has_no_children:
    if "children" in actual_child:
        expected_child["children"] = actual_children
    else:
        expected_child.pop("children", None)
    return
```

PR #708, opened 2026-04-15
Add test for idempotent errors

MaxAake:
Closes: DRIVERS-322


nutkit/frontend/driver.py
```diff
@@ -69,6 +70,7 @@ def __init__(self, backend, uri, auth_token, user_agent=None,
             telemetry_disabled=telemetry_disabled,
             client_certificate=client_certificate_,
             client_certificate_provider_id=client_certificate_provider_id_,
+            disable_auto_commit_retries=disable_auto_commit_retries
```
```suggestion
            disable_auto_commit_retries=disable_auto_commit_retries,
```

nutkit/frontend/driver.py
```diff
@@ -200,15 +202,17 @@ def session(self, access_mode, bookmarks=None, database=None,
...
         req = protocol.NewSession(
             self._driver.id, access_mode, bookmarks=bookmarks,
             database=database, fetchSize=fetch_size,
             impersonatedUser=impersonated_user,
             bookmark_manager=bookmark_manager,
             auth_token=auth_token,
             notifications_min_severity=notifications_min_severity,
-            notifications_disabled_categories=notifications_disabled_categories
+            notifications_disabled_categories=notifications_disabled_categories,  # noqa: E501
+            disable_auto_commit_retries=disable_auto_commit_retries
```
```suggestion
            disable_auto_commit_retries=disable_auto_commit_retries,
```

nutkit/protocol/requests.py
```diff
@@ -78,6 +78,7 @@ def __init__(
         notifications_disabled_categories=None,
         telemetry_disabled=None,
         client_certificate=None, client_certificate_provider_id=None,
+        disable_auto_commit_retries=None
```
```suggestion
        disable_auto_commit_retries=None,
```

tests/stub/idempotent_retries/scripts/idempotent_error.script
```diff
@@ -0,0 +1,15 @@
+!: BOLT 6.0
+
+C: HELLO {"{}": "*"}
+S: SUCCESS {"server": "Neo4j/2026.0.0", "connection_id": "bolt-1"}
+*: RESET
+A: LOGON {"{}": "*"}
+C: RUN {"U": "*"} {"{}": "*"} {"{}": "*"}
+S: FAILURE {"neo4j_code": "Neo.ClientError.MadeUp.Idempotent", "message": "Something went wrong, but nothing's broken", "diagnostic_record": { "_idempotent": true} }
```
robsdedude:
```suggestion
S: FAILURE {"neo4j_code": "Neo.ClientError.MadeUp.Idempotent", "message": "Something went wrong, but nothing's broken", "diagnostic_record": {"_idempotent": true}}
```

tests/stub/idempotent_retries/scripts/idempotent_error.script
```diff
@@ -0,0 +1,15 @@
+!: BOLT 6.0
+
+C: HELLO {"{}": "*"}
+S: SUCCESS {"server": "Neo4j/2026.0.0", "connection_id": "bolt-1"}
```
robsdedude: Is there a reason for this not to be

```
A: HELLO {"{}": "*"}
```

?

Same question holds for other scripts.

tests/stub/idempotent_retries/scripts/idempotent_error_followed_by_regular_error.script
```diff
@@ -0,0 +1,23 @@
+!: BOLT 6.0
+
+C: HELLO {"{}": "*"}
+S: SUCCESS {"server": "Neo4j/2026.0.0", "connection_id": "bolt-1"}
+*: RESET
+A: LOGON {"{}": "*"}
+C: RUN {"U": "*"} {"{}": "*"} {"{}": "*"}
+S: FAILURE {"neo4j_code": "Neo.ClientError.MadeUp.Idempotent", "message": "Something went wrong, but nothing's broken", "diagnostic_record": { "_idempotent": true} }
```
robsdedude:
```suggestion
S: FAILURE {"neo4j_code": "Neo.ClientError.MadeUp.Idempotent", "message": "Something went wrong, but nothing's broken", "diagnostic_record": {"_idempotent": true}}
```

tests/stub/idempotent_retries/scripts/idempotent_error.script
robsdedude: Might be worth to misspell or otherwise replace "Idempotent" in the status code. Just to make sure drivers don't end up relying on this bit for determining what to retry on.

Wouldn't have to be for each test, but having at least one like that would be valuable.

tests/stub/idempotent_retries/scripts/idempotent_error_on_telemetry.script
robsdedude, line 1: `tests/stub/idempotent_retries/scripts/idempotent_error.script` is also an idempotent error on run... Maybe there's a better name for this file.

The same concerns the test name `test_retries_idempotent_error_on_session_run`, too.

tests/stub/idempotent_retries/scripts/idempotent_error_on_run.script
```diff
@@ -0,0 +1,21 @@
...
+S: FAILURE {"neo4j_code": "Neo.ClientError.MadeUp.Idempotent", "message": "Something went wrong, but nothing's broken", "diagnostic_record": { "_idempotent": true} }
+{?
+    # For drivers that pipeline a PULL after RUN
+    C: PULL {"n": {"Z": "*"}}
+    S: IGNORED
+?}
+?: RESET
+C: RUN {"U": "*"} {"{}": "*"} {"{}": "*"}
+S: SUCCESS {"fields": ["n"], "qid": 1}
+C: PULL {"n": 1000}
```
robsdedude: No need to test the driver's default fetch size.

```suggestion
C: PULL {"n": {"Z": "*"}}
```

tests/stub/idempotent_retries/scripts/idempotent_error_on_run_with_telemetry.script
```diff
@@ -0,0 +1,23 @@
+!: BOLT 6.0
+
+C: HELLO {"{}": "*"}
+S: SUCCESS {"server": "Neo4j/2026.0.0", "connection_id": "bolt-1", "hints": {"telemetry.enabled": true}}
+*: RESET
+A: LOGON {"{}": "*"}
+C: TELEMETRY {"Z": "2"}
+S: SUCCESS {}
```
robsdedude:
```suggestion
A: TELEMETRY *
```
No need to test telemetry values. Alternatively `{"Z": "*"}`.

Side note: `2` is equivalent to `{"Z": "2"}`. The stubservers' JSON parsers make a difference between `2.0` and `2` even though that's not quite JSON conform 😉

tests/stub/idempotent_retries/scripts/idempotent_error_followed_by_regular_error.script
```diff
@@ -0,0 +1,23 @@
...
+{?
+    # For drivers that pipeline a PULL after RUN
+    C: PULL {"n": {"Z": "*"}}
+    S: IGNORED
+?}
+?: RESET
```
robsdedude:
```suggestion
+: RESET
```

Mandatory `RESET` after `FAILURE`.

tests/stub/idempotent_retries/scripts/idempotent_error_on_run.script
```diff
@@ -0,0 +1,21 @@
...
+*: RESET
+A: LOGON {"{}": "*"}
+C: RUN {"U": "*"} {"{}": "*"} {"{}": "*"}
+S: FAILURE {"neo4j_code": "Neo.ClientError.MadeUp.Idempotent", "message": "Something went wrong, but nothing's broken", "diagnostic_record": { "_idempotent": true} }
+{?
+    # For drivers that pipeline a PULL after RUN
+    C: PULL {"n": {"Z": "*"}}
+    S: IGNORED
+?}
+?: RESET
```
robsdedude:
```suggestion
+: RESET
```

Mandatory `RESET` after `FAILURE`.

tests/stub/idempotent_retries/scripts/idempotent_error_on_run_with_telemetry.script
```diff
@@ -0,0 +1,23 @@
...
+C: RUN {"U": "*"} {"{}": "*"} {"{}": "*"}
+S: FAILURE {"neo4j_code": "Neo.ClientError.MadeUp.Idempotent", "message": "Something went wrong, but nothing's broken", "diagnostic_record": { "_idempotent": true} }
+{?
+    # For drivers that pipeline a PULL after RUN
+    C: PULL {"n": {"Z": "*"}}
+    S: IGNORED
+?}
+?: RESET
```
robsdedude:
```suggestion
+: RESET
```

Mandatory `RESET` after `FAILURE`.

tests/stub/idempotent_retries/scripts/two_idempotent_errors_on_run.script
```diff
@@ -0,0 +1,23 @@
...
+*: RESET
+A: LOGON {"{}": "*"}
+C: RUN {"U": "*"} {"{}": "*"} {"{}": "*"}
+S: FAILURE {"neo4j_code": "Neo.ClientError.MadeUp.Idempotent", "message": "Something went wrong, but nothing's broken", "diagnostic_record": { "_idempotent": true} }
+{?
+    # For drivers that pipeline a PULL after RUN
+    C: PULL {"n": {"Z": "*"}}
+    S: IGNORED
+?}
+?: RESET
```
robsdedude:
```suggestion
+: RESET
```

Mandatory `RESET` after `FAILURE`.

tests/stub/idempotent_retries/test_idempotent_retries.py
```diff
@@ -0,0 +1,142 @@
...
+    TestkitTestCase,
+)
+from tests.stub.shared import StubServer
+
+
+class TestIdempotentRetries(TestkitTestCase):
+
+    required_features = (
+        types.Feature.BOLT_6_0,
+        types.Feature.IDEMPOTENT_RETRIES
```
robsdedude:
```suggestion
        types.Feature.IDEMPOTENT_RETRIES,
```

tests/stub/idempotent_retries/test_idempotent_retries.py
```diff
@@ -0,0 +1,142 @@
...
+    )
+
+    def setUp(self):
+        super().setUp()
+        self._server = StubServer(9001)
+        uri = "bolt://%s" % self._server.address
+        self._driver = Driver(self._backend, uri,
```
robsdedude: Driver is never close.

Also, I see mixed idioms here. Most tests but the one(s) with sub tests us the instance wide driver whereas the sub tests manage their own driver instances leaving the object wide driver being created (and ideally also destroyed) without it ever being used.

Suggested solution:

```python
def _driver(self):
    return Driver(self._backend, ...)
```

and using the driver's context manager in each test instead of having a driver tacked to the test class.

tests/stub/idempotent_retries/test_idempotent_retries.py
```diff
@@ -0,0 +1,142 @@
...
+        self._server.reset()
+        super().tearDown()
+
+    def test_retries_idempotent_error(self):
+        self._server.start(
+            path=self.script_path("idempotent_error_on_run.script")
+        )
+        with self._driver.session("r") as session:
+            result = session.run("RETURN 1 AS n")
+            records = result.list()
```
robsdedude: Consider

```suggestion
            records = list(result)
```

which will internally call `result.next()` in a loop until no records are left. I suggest this because `.list()` is a special TestKit message that represents a certain usage pattern/driver API and has a dedicated feature flag `API_RESULT_LIST`.

Applies to other tests as well, won't comment there.

tests/stub/idempotent_retries/test_idempotent_retries.py
```diff
@@ -0,0 +1,142 @@
...
+            self.assertEqual(types.Record(values=[types.CypherInt(1)]),
```
robsdedude: Definitely worth adding a check that only one Record came back. Especially since we expect the driver to do a loop-dee-loop.

tests/stub/idempotent_retries/test_idempotent_retries.py
robsdedude, line 1: Suggested tests to add:

 * driver does NOT retry an idempotent error coming back as a response to different messages but `RUN`
   * `TELEMETRY`
   * `PULL`
  * This might be too much, but what about testing that explicit transactions do not retry idempotent errors?

tests/stub/idempotent_retries/test_idempotent_retries.py
```diff
@@ -0,0 +1,142 @@
...
+                "idempotent_error_on_run_with_telemetry.script"
+            )
+        )
+        with self._driver.session("r") as session:
+            result = session.run("RETURN 1 AS n")
+            records = result.list()
+            self.assertEqual(types.Record(values=[types.CypherInt(1)]),
+                             records[0])
+            telemetry_requests = self._server.get_requests("TELEMETRY")
+            self.assertEqual(len(telemetry_requests), 1)
```
robsdedude: * Doesn't hurt, but is redundant because the script enforces exactly one `TELEMETRY` message.
  * There is `server.count_requests()`

tests/stub/idempotent_retries/test_idempotent_retries.py
```diff
@@ -0,0 +1,142 @@
...
+            with self.assertRaises(types.DriverError) as exc:
+                # TODO: remove this block once all languages work
+                if get_driver_name() in ["javascript", "dotnet"]:
+                    session.run("RETURN 1 AS n").next()
+                else:
+                    session.run("RETURN 1 AS n")
+            self.assertEqual(exc.exception.code, "Neo.ClientError.MadeUp.Code")
+        self._server.done()
+
+    def test_session_config_reenables_retries(self):
```
robsdedude: Misleading name. The test appears to cover about all combinations of driver and server config options.

tests/stub/idempotent_retries/test_idempotent_retries.py
```diff
@@ -0,0 +1,142 @@
...
+            (None, False, True),
+            (None, True, False),
+            (False, None, True),
+            (False, False, True),
+            (False, True, False),
+        ):
+            with self.subTest(
+                driver_config=driver_config,
+                session_config=session_config,
+                should_retry=should_retry
```
robsdedude:
```suggestion
                should_retry=should_retry,
```

tests/stub/idempotent_retries/test_idempotent_retries.py
```diff
@@ -0,0 +1,142 @@
...
+            ):
+                if should_retry:
+                    self._server.start(
+                        path=self.script_path("idempotent_error_on_run.script")
+                    )
+                else:
+                    self._server.start(
+                        path=self.script_path("idempotent_error.script")
+                    )
```
robsdedude: Nit: slightly nicer to read, I find
```suggestion
                if should_retry:
                    script = "idempotent_error_on_run.script"
                else:
                    script = "idempotent_error.script"
                self._server.start(path=self.script_path(script))
```

tests/stub/idempotent_retries/test_idempotent_retries.py
```diff
@@ -0,0 +1,142 @@
...
+                                records[0]
+                            )
+                        else:
+                            with self.assertRaises(types.DriverError) as exc:
+                                # TODO: remove this block once js and .net work
+                                if get_driver_name() in [
+                                    "javascript",
+                                    "dotnet"
+                                ]:
+                                    session.run("RETURN 1 AS n").next()
```
robsdedude: The indentation level approaches preposterous/10 😉 Might be worth to move some of this into a function or such to help with this.

tests/stub/idempotent_retries/scripts/idempotent_error.script
MaxAake: Used to be I had telemetry enabled for all tests, when taking that away I neglected to revert to an `A:`

tests/stub/idempotent_retries/scripts/idempotent_error.script
MaxAake: Makes sense to me!

tests/stub/idempotent_retries/test_idempotent_retries.py
MaxAake: indeed, I forgot to rename this one!

tests/stub/idempotent_retries/test_idempotent_retries.py
MaxAake: Ah! Yes that is indeed better

tests/stub/idempotent_retries/test_idempotent_retries.py
MaxAake: That's true, but I like being explicit here, makes what exactly the test is testing clear

tests/stub/idempotent_retries/test_idempotent_retries.py
MaxAake: Made a slightly different version of the function, wich aligns with how the `_driver()` is made in `test_unsupported_type` and `test_vector_types`.

tests/stub/idempotent_retries/test_idempotent_retries.py
MaxAake: Doing this in fact showed that every single test here is covered by this exact code, extracting into a function greatly reduced repeated code.

tests/stub/idempotent_retries/scripts/idempotent_error_on_pull.script
```diff
@@ -0,0 +1,11 @@
+!: BOLT 6.0
+
+A: HELLO {"{}": "*"}
+*: RESET
+A: LOGON {"{}": "*"}
```
robsdedude:
```suggestion
A: LOGON {"{}": "*"}
*: RESET
```
same for all other scripts. The server does not allow you to `RESET` in the connection when unauthenticated.

tests/stub/idempotent_retries/test_idempotent_retries.py
```diff
@@ -0,0 +1,184 @@
...
+            self._backend, uri,
+            types.AuthorizationToken(
+                "basic", principal="", credentials=""
+            ),
+            disable_auto_commit_retries=disable_auto_commit_retries,
+        )
+        try:
+            yield driver
+        finally:
+            driver.close()
```
robsdedude: `Driver` already implements the contextmanager protocol (by implementing the magic methods `__enter__` and `__exit__`). So this could be simplified (and actually improved because `Driver.__exit__` tries to improve error propagation)

```suggestion
    @contextmanager
    def _driver(self, disable_auto_commit_retries=None):
        uri = "bolt://%s" % self._server.address
        auth = types.AuthorizationToken("basic", principal="", credentials="")
        with Driver(
            self._backend, uri, auth,
            disable_auto_commit_retries=disable_auto_commit_retries,
        ) as driver:
            yield driver
```

or even shorter if you don't mind exposing the whole driver and not just a context manager (thus forcing tests to use a `with` statement)

```suggestion
    def _driver(self, disable_auto_commit_retries=None):
        uri = "bolt://%s" % self._server.address
        auth = types.AuthorizationToken("basic", principal="", credentials="")
        return Driver(
            self._backend, uri, auth,
            disable_auto_commit_retries=disable_auto_commit_retries,
        )
```

tests/stub/idempotent_retries/test_idempotent_retries.py
```diff
@@ -0,0 +1,184 @@
...
+                        records[0]
+                    )
+                    self.assertEqual(1, len(records))
+                else:
+                    with self.assertRaises(types.DriverError) as exc:
+                        # TODO: remove this block once js and .net work
+                        if get_driver_name() in [
+                            "javascript",
+                            "dotnet"
+                        ] or error_on_pull:
```
robsdedude: Comment is outdated. Not the whole block needs to go. Maybe something like this

```suggestion
                        # TODO: remove this block once js and .net work
                        if get_driver_name() in ["javascript", "dotnet"]:
                            error_on_pull = True
                        if error_on_pull:
```

or just rewording the comment

tests/stub/idempotent_retries/scripts/idempotent_error_on_pull.script
```diff
@@ -0,0 +1,11 @@
...
+C: RUN {"U": "*"} {"{}": "*"} {"{}": "*"}
+S: SUCCESS {"fields": ["n"], "qid": 1}
+C: PULL {"n": {"Z": "*"}}
+S: FAILURE {"neo4j_code": "Neo.ClientError.MadeUp.Idempotent", "message": "Something went wrong, but nothing's broken", "diagnostic_record": { "_idempotent": true} }
+*: RESET
```
robsdedude:
```suggestion
+: RESET
```
It's the end of the test, so I guess the driver would be fine to just ditch the connection... But do we really expect that? At the time of the FAILURE the driver does not yet know that it will be closed shortly... So I think it's appropriate to expect a `RESET`.

tests/stub/idempotent_retries/scripts/idempotent_error_on_telemetry.script
```diff
@@ -0,0 +1,20 @@
...
+{?
+    C: RUN {"U": "*"} {"{}": "*"} {"{}": "*"}
+    S: IGNORED
+?}
+{?
+
+    # For drivers that pipeline a PULL after RUN
+    C: PULL {"n": {"Z": "*"}}
+    S: IGNORED
+?}
```
robsdedude, line 17:
```suggestion
{?
    C: RUN {"U": "*"} {"{}": "*"} {"{}": "*"}
    S: IGNORED
    {?
        # For drivers that pipeline a PULL after RUN
        C: PULL {"n": {"Z": "*"}}
        S: IGNORED
    ?}
?}
```
Not really what we're testing here, but doesn't hurt to rule out a message order that a real server would def. not accept in a non-failure mode (i.e., `PULL` without `RUN` beforehand).

tests/stub/idempotent_retries/test_idempotent_retries.py
robsdedude: Good point :heavy_check_mark:

tests/stub/idempotent_retries/test_idempotent_retries.py
robsdedude: https://github.com/neo4j-drivers/testkit/pull/708/changes#r3122306657

tests/stub/idempotent_retries/test_idempotent_retries.py
MaxAake: Interesting! Might be worth carding up changing the `_driver()` functions on `test_vector_types.py` and `test_unsupported_type.py` then too, if only to avoid someone copying the logic in them in the future

robsdedude: Maybe. To be fair the whole of TestKit could use some refactoring love of this kind. There are plenty of tests that don't use any kind of context manager, not even a `try/finally` such that failing tests cascade into more fires by not closing resources... There's a fair bit of tech debt in this testing framework :grimacing:

robsdedude (approved):
➰ 🧼

PR #712, opened 2026-05-12
Avoid race condition on which error is reported for the JS driver

MaxAake:
Part of: DRIVERS-309

Due to similar reasons that JS and dotnet have to call next() on `session.run()` results to get an error, not consuming the result of `tx.run()` triggers a race condition in the JS backend, leaving it up to timing if the correct error gets reported or if it's an error relating to committing a failed transaction.


robsdedude (approved):
🏖️🏐

PR #720, opened 2026-06-11
HTTP/Query API tests

robsdedude:
Closes: DRIVERS-326


tests/stub/http_query/shared/http_endpoint_builders/_tx.py
```diff
@@ -0,0 +1,435 @@
...
+            HttpTxRollbackEndpoint.RequestData(
+                db=self._db, tx_id=self._tx_id, auth=self._auth
+            ),
+            HttpTxRollbackEndpoint.ResponseData(
+                legacy=legacy_response,
+                errors=errors,
+                status_code=status_code,
+            ),
+        )
+        return self
```
StephenCathcart, line 262: `with_query` passes `_verify_affinity_header` in but commit and rollback don't, though both endpoint classes take it. fed the matchers a request each way:

    query endpoint   without the header -> no match
    commit endpoint  without the header -> still matches

so `test_affinity_header_*` passes either way. docs say it's needed on commit/rollback too, and a real server 404s the commit without it. added the check to both and the suite still passed

```suggestion
    def with_commit(
        self,
        bookmarks: list[str] | None = None,
        errors: list[dict[str, object]] | None = None,
        status_code: int | None = None,
    ) -> t.Self:
        self._finishing_handler = HttpTxCommitEndpoint(
            HttpTxCommitEndpoint.RequestData(
                db=self._db, tx_id=self._tx_id, auth=self._auth
            ),
            HttpTxCommitEndpoint.ResponseData(
                bookmarks=bookmarks, errors=errors, status_code=status_code
            ),
            extra_header_verification=(self._verify_affinity_header,),
        )
        return self

    def with_rollback(
        self,
        *,
        legacy_response: bool = False,
        errors: list[dict[str, object]] | None = None,
        status_code: int | None = None,
    ) -> t.Self:
        self._finishing_handler = HttpTxRollbackEndpoint(
            HttpTxRollbackEndpoint.RequestData(
                db=self._db, tx_id=self._tx_id, auth=self._auth
            ),
            HttpTxRollbackEndpoint.ResponseData(
                legacy=legacy_response,
                errors=errors,
                status_code=status_code,
            ),
            extra_header_verification=(self._verify_affinity_header,),
        )
        return self
```

tests/stub/http_query/test_retries.py
```diff
@@ -0,0 +1,154 @@
...
+QUERY = "RETURN 1 AS n"
+FIELDS = ["n"]
+RECORDS: list[list[http_types.HttpType]] = [[http_types.Int(1)]]
+
+
+class _FailPoint(enum.Enum):
+    PRE_TX_CREATION = enum.auto()
+    PRE_QUERY_HEADER = enum.auto()
+    PRE_QUERY_RECORDS = enum.auto()
+    POST_QUERY_RECORDS = enum.auto()
```
StephenCathcart, line 34: this happens before the query finishes, so the driver always gets a proper error body and knows the commit didn't land. couldn't see anything for a commit that gets no answer at all

`test_auth_token_manager` does put errors in `with_commit` / `with_rollback`, but that's a response rather than a dropped connection. looks like you had in the `TODO.md` "amend: tx commit & rollback" unticked so you might already be aware?

bolt has `test_retry.py::test_disconnect_on_commit` for this. works here too with a `drop_connection` flag on `HttpTxCommitEndpoint.ResponseData` that streams a body shorter than its `Content-Length`. tried it and it caught a real retry:

    FAIL: test_no_retry_on_commit_disconnect
    AssertionError: DriverError not raised

can you double check this? im assuming it wants the java/dotnet skip that the bolt one has

tests/stub/http_query/shared/http_endpoints/_query.py
```diff
@@ -0,0 +1,362 @@
...
+        plan: Plan | None = None
+        profile: Profile | None = None
+        notifications: list[Notification] | None = None
+        errors: list[dict[str, object]] | None = None
+        status_code: int | None = None
+
+        def _get_counters(self, req: Request) -> t.Any:
+            if isinstance(self.counters, AutoRespond):
+                body = req.get_json(force=True)
+                if body.get("include_counters") is True:
```
StephenCathcart, line 162: i think the real field is `includeCounters`, so `AutoRespond` never fires. same in `_tx_query.py:135` and `_tx.py:177`, which also read `query` instead of `statement` (their own matchers below use `statement`).

    before: includeCounters: true  -> no counters
    after:  includeCounters: true  -> counters
    after:  includeCounters absent -> no counters

probs not a problem right now since the counters tests pass an explicit `CountersMap`, but might be an issue later? fixed all three and the suite still passed.

```suggestion
                if body.get("includeCounters") is True:
```

tests/stub/http_query/test_errors.py
```diff
@@ -0,0 +1,666 @@
...
+        .build()
+    )
+
+
+STATUS_CODES = [
+    None,  # auto
+    # Some early server versions sometimes return errors with 2xx status codes.
+    # Drivers should pick up the error anyway and throw.
+    200,
+    400,
```
StephenCathcart, line 112: had a poke at a 2025.04.0 server to see what it sends with errors: runtime ones (`RETURN 1/0`, or after some records) are 202 with `data` + `errors`, syntax errors and bad bodies are 400, missing db / unknown tx are 404. same inside a tx.

so 400's fine, but the default error here is `ArithmeticError` "/ by zero" which is a 202 case, and 202 isn't in the list. added it and `test_errors` still passed. does 200 cover it fine?

```suggestion
STATUS_CODES = [
    None,  # auto
    # Some early server versions sometimes return errors with 2xx status codes.
    # Drivers should pick up the error anyway and throw.
    200,
    # What current servers return for errors raised during execution.
    202,
    400,
```

tests/stub/http_query/shared/http_endpoints/_base.py
```diff
@@ -0,0 +1,351 @@
...
+        if protocol_version == ProtocolVersion.V1_0:
+            return {
+                "Content-Type": "application/vnd.neo4j.query",
+            }
+        elif protocol_version == ProtocolVersion.V1_1:
+            return {
+                "Content-Type": "application/vnd.neo4j.query.v1.1",
+            }
+        else:
+            NotImplementedError("TODO")
```
StephenCathcart, line 318: do the errors need raised?

```suggestion
        if protocol_version == ProtocolVersion.V1_0:
            return content_type in {
                "application/vnd.neo4j.query",
                "application/vnd.neo4j.query.v1.0",
            }
        elif protocol_version == ProtocolVersion.V1_1:
            return content_type == "application/vnd.neo4j.query.v1.1"
        else:
            raise NotImplementedError("TODO")

    @classmethod
    def _version_as_header(
        cls,
        protocol_version: ProtocolVersion,
    ) -> dict[str, str]:
        if protocol_version == ProtocolVersion.V1_0:
            return {
                "Content-Type": "application/vnd.neo4j.query",
            }
        elif protocol_version == ProtocolVersion.V1_1:
            return {
                "Content-Type": "application/vnd.neo4j.query.v1.1",
            }
        else:
            raise NotImplementedError("TODO")
```

tests/stub/http_query/shared/http_endpoints/_base.py
```diff
@@ -0,0 +1,351 @@
...
+    "serialize_any",
+    "TestKitRequestMatcher",
+    "url_encode",
+)
+
+
+@dataclasses.dataclass
+class MaybeNull(t.Generic[T]):
+    """
+    Instruct a matcher to accept requests with this value missing/bein None.
```
StephenCathcart, line 54:
```suggestion
    Instruct a matcher to accept requests with this value missing/being None.
```

tests/stub/http_query/shared/http_endpoint_builders/_tx.py
```diff
@@ -0,0 +1,435 @@
...
+
+    def without_finishing_handler(self) -> t.Self:
+        self._finishing_handler = None
+        return self
+
+    def build(self) -> HttpEndpoint:
+        assert self._sequential_handlers, "initialized non-empty int __init__"
```
StephenCathcart, line 269:
```suggestion
        assert self._sequential_handlers, "initialized non-empty in __init__"
```

StephenCathcart (approved):
Mainly open questions - looks good! 👍

tests/stub/http_query/test_errors.py
robsdedude: This is a good question. I was quietly assuming that all drivers only ever look at status codes in 100-bins (2xx-ok, 3xx-redirect, 4xx-client error, 5xx-server error) for programmatic handling inside the driver code (besides maybe different error messages). Maybe I shouldn't have done that. In such case, I suggest we also add 404 as you said real servers respond with 404 as well.

tests/stub/http_query/shared/http_endpoints/_base.py
robsdedude: 🤦️

tests/stub/http_query/test_errors.py
StephenCathcart: Yeah I think returning 404 would be a good idea, it might save us some edge cases or work later on, but also don't want to add any redundant code if its "just in case" - if it's not much more effort I'd add the the extra codes 👍

StephenCathcart (approved):
Looks good! 👍

PR #722, opened 2026-06-13
Add neo4j integration tests for UUID type (Bolt 6.1)

RichardIrons-neo4j:
## Summary

Adds integration tests for the UUID type introduced in Bolt 6.1, running against a real Neo4j instance.

**`tests/neo4j/datatypes/test_uuid.py`:**
- `test_should_echo_uuid` — scalar UUIDs (nil, all-ones, sequential, random) sent as parameters and echoed back
- `test_uuid_in_list` — list of UUIDs round-trips correctly
- `test_uuid_in_map` — UUID as a map value
- `test_cypher_created_uuid` — server-generated UUID via `uuid()` returns a native `CypherUUID`
- `test_uuid_stored_on_node` — UUID written and read back as a node property

All tests require `Feature:API:Type.UUID` and `Feature:Bolt:6.1`.

Closes: DRIVERS-441

### DIFF
--- a/tests/neo4j/datatypes/test_uuid.py
+++ b/tests/neo4j/datatypes/test_uuid.py
@@ -1,0 +1,55 @@
+import uuid
+
+import nutkit.protocol as types
+from tests.neo4j.datatypes._base import _TestTypesBase
+
+
+class TestUuidTypes(_TestTypesBase):
+
+    required_features = (
+        types.Feature.API_TYPE_UUID,
+        types.Feature.BOLT_6_1,
+    )
+
+    def test_should_echo_uuid(self):
+        values = [
+            uuid.UUID("00000000-0000-0000-0000-000000000000"),  # nil UUID
+            uuid.UUID("ffffffff-ffff-ffff-ffff-ffffffffffff"),  # all-ones
+            uuid.UUID("01020304-0506-0708-090a-0b0c0d0e0f12"),  # sequential bytes
+            uuid.uuid4(),
+        ]
+        self._create_driver_and_session()
+        for value in values:
+            with self.subTest(value=str(value)):
+                self._verify_can_echo(types.CypherUUID(value))
+
+    def test_uuid_in_list(self):
+        self._create_driver_and_session()
+        data = types.CypherList([
+            types.CypherUUID(uuid.UUID("00000000-0000-0000-0000-000000000000")),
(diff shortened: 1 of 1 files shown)


MaxAake (changes requested):
The tests look good, but they need a tag to ensure they are only run when the server supports Bolt 6.1. Something similar to `requires_vector_support` in `shared.py`

RichardIrons-neo4j:
> The tests look good, but they need a tag to ensure they are only run when the server supports Bolt 6.1. Something similar to `requires_vector_support` in `shared.py`

I was hoping lines 9-12 in the new file would handle that?

MaxAake:
Those will only ensure that the driver supports bolt 6.1 and UUIDs, not that it is communicating with a server version that matches. This is why the Java tests are failing (I think Java is the only driver with UUID merged)

RichardIrons-neo4j:
> Those will only ensure that the driver supports bolt 6.1 and UUIDs, not that it is communicating with a server version that matches. This is why the Java tests are failing (I think Java is the only driver with UUID merged)

Cool, didn't realise - thanks! I've updated the file.

MaxAake (approved):
Looks good!

I leave for you to decide if you want to merge now or let this wait until it can be tested against a nightly build with Bolt 6.1 enabled.

RichardIrons-neo4j:
I'm going to be bold and merge it.

tests/neo4j/datatypes/test_uuid.py
```diff
@@ -0,0 +1,63 @@
...
+        value = uuid.uuid4()
+        self._create_driver_and_session()
+        self._verify_can_echo(
+            types.CypherMap({"id": types.CypherUUID(value)})
+        )
+
+    @requires_min_bolt_version("6.1")
+    def test_cypher_created_uuid(self):
+        self._create_driver_and_session()
+        values = self._read_query_values("RETURN uuid()")
```
robsdedude, line 51: Great PR 👏 

I feel like this test could be improved. If the server-side UUID wasn't random, but created with a specific value, the test could check that the value is truthfully returned by the driver. Because all these tests (not the stub tests though) would pass even if the driver were to say wrap-subtract 1 from each byte when sending and then wrap-add 1 to each byte when receiving a UUID.

robsdedude: DRIVERS-460

PR #727, opened 2026-08-05
Testkit tests for property encryption

RichardIrons-neo4j:
# Property encryption

Feature flag: `Feature.API_PROPERTY_ENCRYPTION`. 

The byte-exact tests (`test_encrypts_to_known_bytes` and `test_decrypts_known_bytes`) share one fixture list in `tests/stub/property_encryption/deterministic_fixtures.py`, asserted in both directions. Every input is pinned there — the KEK, the encapsulated DEK and its wrap metadata, the key id, and each case's value, IV and AAD — so the expected bytes depend only on those inputs and not on any backend's key generation or id assignment. The fixtures were produced with the .NET backend; a driver that disagrees is a discrepancy worth investigating rather than a fixture to regenerate.

## Protocol additions

Throughout this doc, `bytes` means a bare space-separated lower-case hex string on the wire (e.g. `"0a 1b 2c"`), the same convention `CypherVector`'s `data` field uses — not a `CypherBytes` Cypher-value envelope. This applies to `kek`, `iv`, `encapsulation`, `encryptedBytes` and `Decrypt`'s `value` below, since none of them are themselves a Cypher property value. `EncryptToBytes`'s `value` and both `aad` fields are the only ones carrying real property values, so they stay full `<any Cypher value>` envelopes.

### NewDriver

`NewDriver` gains one optional field:

```
propertyEncryptionProfiles: [{ name: string, kek: bytes | null }]
```

Each entry configures one property-encryption profile on the driver. The profile is of the Envelope kind, backed by a backend-owned fixture `KeyEncapsulationService`/`EncapsulatedKeyRepository` pair; the backend plays the role of the user, who in a real application owns and supplies these objects directly. Profile names must be unique within a driver. If the field is omitted, no encryption profiles are configured.

`kek`, when given, constructs the profile's fixture key-encapsulation service with that exact key instead of a random one. This exists solely to support the deterministic and cross-driver decrypt tests; ordinary tests omit it.

#### Fixture key-encapsulation service

The fixture `KeyEncapsulationService` is backend-owned, but its wrap format is a cross-driver contract, not a free choice: the deterministic fixtures ship a pre-made encapsulation with its metadata, so every backend has to unwrap what another one produced. Implement it as

- AES-256-GCM, with the profile's `kek` as the key-encryption key (or a random 32-byte key when `kek` is omitted),
- a 32-byte data-encryption key,
- a fresh 12-byte IV per encapsulation, and no AAD,
- a 16-byte tag,
- `encapsulation` = ciphertext followed by tag, so 48 bytes for a 32-byte DEK,
- the IV returned as encapsulation metadata, base64-encoded under the key `iv`.

A backend that differs here still passes the round-trip tests, which never cross backends, and then fails the deterministic ones with nothing to point at the cause.

### EncryptToBytes

```
EncryptToBytes {
  driverId: string
  value: <any Cypher value>
  aad: <any Cypher value> | null
  profileName: string | null   # null => use the sole configured profile
  keyAlias: string | null      # exactly one of keyAlias/keyId must be set
  keyId: string | null
  iv: bytes | null             # null => random IV as normal
}
-> EncryptedValue { encryptedBytes: bytes }
-> Error
```

Encrypts `value`, binding `aad` when supplied. `profileName` may be omitted only when exactly one profile is configured; with several configured and none named, the driver raises.

`iv`, when given, is the exact 12-byte IV the driver must use for this one encrypt call, and a backend should reject one that is not 12 bytes. With the data-encryption key pinned via `ImportEncapsulatedKey` and the IV pinned here, an encrypt is fully deterministic, so tests can assert byte-exact ciphertext instead of only round-tripping. Like `kek`, this exists solely for tests that ask for determinism; ordinary tests omit it. `test_encrypts_to_known_bytes` uses it to check the driver's encryption output byte-for-byte against the fixtures described above.

A backend is not expected to verify that a pinned IV was actually consumed. Whether a given request goes on to perform an encrypt is testkit's business, not the backend's; a backend that treats an unconsumed IV as an error turns one genuine failure inside a loop of pinned-IV cases into a run of misleading ones.

### Decrypt

```
Decrypt {
  driverId: string
  value: bytes                 # as returned by EncryptToBytes
  aad: <any Cypher value> | null
  usePersistedAad: bool        # exactly one of aad/usePersistedAad must be set
}
-> DecryptedValue { decryptedValue: <any Cypher value> }
-> Error
```

The profile is resolved from the encrypted bytes: each encrypted value records which profile produced it, so the caller does not name one.

### CreateEncapsulatedKey

```
CreateEncapsulatedKey {
  driverId: string
  alias: string
  profileName: string | null   # null => use the sole configured profile
}
-> EncapsulatedKey { id: string, alias: string }
-> Error
```

Generates a fresh data-encryption key through the profile's `KeyEncapsulationService` and registers it in the profile's `EncapsulatedKeyRepository` under `alias`. A key must exist under a given alias before that alias can be used for encryption.

The response deliberately carries only the id and alias. The encapsulation and its key-encapsulation metadata are not the driver's to report back: the caller supplied the `KeyEncapsulationService` and `EncapsulatedKeyRepository` that produced and stored them, so a caller needing those bytes reads its own repository.

### ImportEncapsulatedKey

```
ImportEncapsulatedKey {
  driverId: string
  id: string                     # the id to store the key under
  alias: string
  encapsulation: bytes           # the wrapped-DEK bytes
  metadata: map<string,string>   # KES metadata for the encapsulation (e.g. the wrap IV)
  profileName: string | null     # null => use the sole configured profile
}
-> EncapsulatedKey { id: string, alias: string }
-> Error
```

Seeds a profile's repository with a pre-made encapsulated key directly, skipping the key-generation step `CreateEncapsulatedKey` performs, since the caller already has an encapsulation from elsewhere (a fixture, or a prior `CreateEncapsulatedKey` response). This corresponds to what a real user would do to import a key obtained elsewhere: call their own repository's save method directly, without going through the driver's `PropertyEncryption` API at all.

`id` is explicit rather than repository-assigned because the key id is recorded in the Encrypted structure's metadata, so it forms part of the encrypted bytes. Leaving it to the backend's own id-generation scheme made the byte-exact fixtures silently depend on that scheme matching across drivers, which it does not. It is named `id` rather than `keyId` to match its unprefixed sibling `alias`; `EncryptToBytes` keeps `keyId`, where the prefix pairs it with `keyAlias`.


tests/stub/property_encryption/test_property_encryption.py
```diff
@@ -0,0 +1,187 @@
...
+    def test_round_trips_many_values_in_shuffled_order(self):
+        driver = self._new_driver()
+        driver.create_encapsulated_key("k1")
+
+        values = [
+            types.CypherBool(True),
+            types.CypherBool(False),
+            types.CypherInt(0),
+            types.CypherInt(-1),
+            types.CypherInt(9223372036854775807),
```
MaxAake, line 63: The javascript driver backend recieves CypherInts as JS numbers, meaning numbers this large cause floating point errors. The JS driver can round trip a MAX_INT64 without problem when the testkit backend isn't involved. Either we skip this test on JS, or, if acceptable, we use a smaller number on this one.

tests/stub/property_encryption/test_property_encryption.py
```diff
@@ -0,0 +1,187 @@
...
+        driver_2.import_encapsulated_key(
+            "k1", key.encapsulated_bytes, key.metadata, profile_name="fx"
+        )
+
+        decrypted = driver_2.decrypt(encrypted, use_persisted_aad=True)
+
+        self.assertEqual(decrypted, types.CypherString("hello world"))
+
+    def test_decrypts_values_produced_by_other_drivers(self):
+        for case in DECRYPT_INTEROP_TEST_CASES:
```
MaxAake, line 205: Haven't been able to get this to work, the decryption job fails in a way that indicates some input is malformed (basically the authentication tag doesn't match). It'd probably be easiest to debug this together

Interestingly the key decapsulation works, so I should be using the same key, and manually deserializing the hexcode of the struct it looks like the data going into the decrypt function is the exact message and IV.

MaxAake (commented):
💬

nutkit/protocol/requests.py
```diff
@@ -891,3 +894,111 @@ class FakeTimeUninstall:
...
+    :param metadata: The key encapsulation service's metadata for the
+        encapsulation.
+    :param profile_name: The name of the encryption profile to import the
+        key into, or None to use the sole configured profile.
+    """
+
+    def __init__(self, driver_id, alias, encapsulation, metadata,
+                 profile_name=None):
+        self.driverId = driver_id
+        self.alias = alias
```
injectives, line 1006: I think we should add the `keyId` here as well as the identifier goes into the metadata of the Encrypted Structure and it must be predictable for stub tests asserting on expected encrypted bytes. The message handler should take all of this in import the key directly into its key repository.

nutkit/protocol/responses.py
```diff
@@ -343,6 +343,42 @@ def __init__(self, id):
...
+    An encapsulated data encryption key.
+
+    Sent in response to a CreateEncapsulatedKey request.
+    """
+
+    def __init__(self, id, alias, encapsulatedBytes, metadata):
+        self.id = id
+        self.alias = alias
+        self.encapsulated_bytes = bytes.fromhex(encapsulatedBytes)
+        self.metadata = metadata
```
injectives: `encapsulatedBytes` and `metadata` are not public API on DEK creation. I think it would be better if we did not require them here.

tests/stub/property_encryption/test_property_encryption.py
```diff
@@ -0,0 +1,241 @@
...
+        )
+        driver_2.import_encapsulated_key(
+            "k1", key.encapsulated_bytes, key.metadata, profile_name="fx"
+        )
+
+        decrypted = driver_2.decrypt(encrypted, use_persisted_aad=True)
+
+        self.assertEqual(decrypted, types.CypherString("hello world"))
+
+    def test_encrypts_to_known_bytes(self):
```
injectives, line 262: @RichardIrons-neo4j, I have been able to make it work with Java Driver, but only when I hardcode `"0"` key id on import. I think addressing the following will resolve this: https://github.com/neo4j-drivers/testkit/pull/727/changes#r3824978418

nutkit/protocol/requests.py
RichardIrons-neo4j: Added.

nutkit/protocol/responses.py
RichardIrons-neo4j: Took these out.

tests/stub/property_encryption/test_property_encryption.py
RichardIrons-neo4j: Fixed by adding key_id to `ImportEncapsulatedKey`. The tests have been recreated using `"testkit-key"` as the key. There are now deterministic tests in both directions against those fixtures.

nutkit/protocol/requests.py
```diff
@@ -891,3 +894,116 @@ class FakeTimeUninstall:
...
+    :param metadata: The key encapsulation service's metadata for the
+        encapsulation.
+    :param profile_name: The name of the encryption profile to import the
+        key into, or None to use the sole configured profile.
+    """
+
+    def __init__(self, driver_id, key_id, alias, encapsulation, metadata,
+                 profile_name=None):
+        self.driverId = driver_id
+        self.keyId = key_id
```
injectives: For consistency with `alias`, you might want to rename this to `id`.

tests/stub/property_encryption/test_property_encryption.py
robsdedude, line 1: Suggest adding tests for types rejected by the encryption function. Such as
 * maps
 * lists of inhomogeneous types
   * especially interesting zoned datetimes/times are considered other types as local datetimes/times
   * likewise are spatial points with different SRIDs
 * list of
   * lists
   * vectors
   * maps

MaxAake: I have added a test that ensure all of these cases throw!

MaxAake (approved):
Looks good and works on current JavaScript Driver ✅ 

I did write the last commit on this PR, though. so it should probably be reviewed by 1 more person

PR #728, opened 2026-08-08
Skip building the Rust boltstub when TEST_RUSTY_STUB isn't set

RichardIrons-neo4j:
The runner image unconditionally installed the Rust toolchain and ran `cargo build --release` regardless of which stub server the run actually uses, using disk and time on every non-rusty-stub job. This was traced back as the root cause of several disk-space CI failures: non-rusty-stub builds only requested 3gb of space, but the toolchain install, `apt-get` packages, and the compile itself happened every time.

Verified on PR #727's full CI run: the javascript composite, which had failed four times in a row on disk exhaustion, passed cleanly once this fix was in place, and all 14 checks on that PR went green.

### DIFF
--- a/runner.py
+++ b/runner.py
@@ -3,6 +3,11 @@ import subprocess
 
 import docker
 
+USE_RUST = (
+    os.environ.get("TEST_RUSTY_STUB", "").lower()
+    in ("true", "y", "yes", "1", "on")
+)
+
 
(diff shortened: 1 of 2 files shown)


RichardIrons-neo4j:
I'm self-merging this since it is a cause of flaky builds across the team.

PR #729, opened 2026-08-17
Remove dotnet skip for retry-clustering tests

RichardIrons-neo4j:
Fixes DRIVERS-511

Removes the three hardcoded dotnet skips ("Behaves strange") from `test_retry_clustering.py`. The rewritten .NET testkit backend handles the retryable-negative flow correctly, and all three tests pass against it (verified locally against the rewrite at its current head).

neo4j/neo4j-dotnet-driver#957 has now merged, so this PR is now safe to merge too.


AndyHeap-NeoTech (approved):

AndyHeap-NeoTech (approved):
Looks good to me.

PR #730, opened 2026-08-18
Remove UUID preview env flags now UUID is GA in 2026.09 nightlies

RichardIrons-neo4j:
The `2026-community-debian-nightly` image is now 2026.09.0, where UUID went GA and `internal.cypher.uuid_type_enabled` was removed. Testkit still passes it, which fails strict config validation, so the server exits about 4 seconds after start, and the run aborts.

I ran the current nightly with no overrides at all: it boots fine, negotiates Bolt 6.1 by default, and `RETURN uuid('...')` round-trips with `valueType` reporting `UUID NOT NULL`. The whole `[uuid-preview]` env block is obsolete, not just the one removed setting, so this deletes both copies (`Standalone.start` and `Core.start`).

The two remaining `[uuid-preview]` comments (the `max_protocol_version` bound in `tests/neo4j/shared.py` and the released-version matrix TODO in `main.py`) need the final GA version number when the UUID stuff actually ships.

### DIFF
--- a/neo4j.py
+++ b/neo4j.py
@@ -70,16 +70,6 @@ class Standalone:
         if self._edition != "community":
             env_map["NEO4J_ACCEPT_LICENSE_AGREEMENT"] = "yes"
 
-        # TODO: remove once UUID is GA
-        # [uuid-preview] search tag for removal of UUID preview workarounds
-        if (2026, 4) < self._version:
-            env_map.update({
-                "NEO4J_internal_dbms_latest__runtime__version": "2147483647",
-                "NEO4J_internal_dbms_latest__kernel__version": "254",
-                "NEO4J_internal_cypher_uuid__type__enabled": "true",
-                "NEO4J_internal_dbms_bolt_max__protocol__version": "6.1",
-            })
-
         logs_path = join(self._artifacts_path, "logs")
         self._container = docker.run(
             self._image, self._hostname,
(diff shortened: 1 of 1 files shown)


bigmontz (dismissed):
🥇

neo4j.py
```diff
@@ -70,16 +70,6 @@ def start(self, network):
...
-        # TODO: remove once UUID is GA
-        # [uuid-preview] search tag for removal of UUID preview workarounds
-        if (2026, 4) < self._version:
-            env_map.update({
-                "NEO4J_internal_dbms_latest__runtime__version": "2147483647",
-                "NEO4J_internal_dbms_latest__kernel__version": "254",
-                "NEO4J_internal_cypher_uuid__type__enabled": "true",
-                "NEO4J_internal_dbms_bolt_max__protocol__version": "6.1",
-            })
-
```
bigmontz, line 82: Don't we test against other versions between 2026.04 and now?

bigmontz (approved):
Just a small comment, but I think if it is fine if we will only test for 2026.09 and following versions.

neo4j.py
RichardIrons-neo4j: No — the matrix only tests released versions up to 5.26 (2025.10 is still commented out pending release), plus nightlies: 5.26 (tag 5) and the 2026.dev configs. There's nothing between 2026.04 and now; the only configs the deleted block ever fired for are the 2026.dev ones, and they all pull the same current nightly, where the UUID flag is now fatal.

PR #732, opened 2026-09-07
Allow the stub suite and the server config list to be sharded across parallel CI jobs

RichardIrons-neo4j:
Two env vars for CI sharding, with no effect if not set.

| Variable | Shards | Applied in |
|---|---|---|
| `TEST_SHARD` | tests within the stub suite | `tests/stub/suites.py` |
| `TEST_CONFIG_SHARD` | the Neo4j server configuration list | `main.py` |

In a .NET driver test run, `tests.stub.routing` is 504 tests of roughly 1200 and ~21% of the entire CI build on its own. Sharding on the basis of test type would create a big slow routing shard and several small cheap ones. This PR allows sharding through round-robin collection of tests.

Doing it by index avoids a hand-written list of modlues somewhere that could risk new tests being missed.

The .NET driver `INTEGRATION_TESTS` build currently takes 43.8 min, of which about 18 min is server boots. Sharding like this reduces the number of servers that need to be booted.

### DIFF
--- a/tests/testenv.py
+++ b/tests/testenv.py
@@ -39,14 +39,18 @@ def parse_shard(spec):
     return index, count
 
 
-def shard_suite(suite, spec):
-    # Tests are taken round-robin so tests from an expansive
-    # module are spread across shards.
+def shard_list(items, spec):
+    # Items are taken round-robin so a run of expensive items is spread
+    # across shards.
     index, count = parse_shard(spec)
+    return items[index - 1::count]
+
+
+def shard_suite(suite, spec):
     tests = list(_flatten(suite))
-    selected = tests[index - 1::count]
+    selected = shard_list(tests, spec)
     print(
-        "Shard %d/%d: running %d of %d tests"
-        % (index, count, len(selected), len(tests))
+        "Shard %s: running %d of %d tests"
+        % (spec.strip(), len(selected), len(tests))
     )
     return unittest.TestSuite(selected)
--- a/tests/testenv.py
+++ b/tests/testenv.py
@@ -26,7 +26,7 @@ def parse_shard(spec):
     parts = [part.strip() for part in spec.split("/")]
     if len(parts) != 2 or not all(part.isdigit() for part in parts):
         raise ValueError(
-            'shard must look like "index/count", for example "2/5", got %r'
+            'shard must look like "index/count", for example "2/5", got "%r"'
             % (spec,)
         )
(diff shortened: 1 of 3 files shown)


tests/testenv.py
```diff
@@ -9,3 +11,46 @@ def get_test_result_class(name):
...
+    if not 1 <= index <= count:
+        raise ValueError(
+            "shard index must be between 1 and %d, got %r" % (count, spec)
+        )
+    return index, count
+
+
+def shard_list(items, spec):
+    # Items are taken round-robin so a run of expensive items is spread
+    # across shards.
```
robsdedude: AFAIK, we currently don't have any test that are using module-wide setup and teardown functions. But this might interfere with such tests if introduced in the future.

More a thing to be aware of than to avoid, I'd say.

RichardIrons-neo4j: Thanks, good note.

tests/testenv.py
```diff
@@ -9,3 +11,41 @@ def get_test_result_class(name):
...
+    parts = [part.strip() for part in spec.split("/")]
+    if len(parts) != 2 or not all(part.isdigit() for part in parts):
+        raise ValueError('shard must be in the form "#/#"')
+
+    index, count = (int(part) for part in parts)
+    if count < 1:
+        raise ValueError("shard count must be at least 1")
+
+    if not 1 <= index <= count:
+        raise ValueError("shard index must be >= 1 and <= %d" % (count,))
```
robsdedude:
```suggestion
        raise ValueError(f"shard index must be >= 1 and <= {count}")
```
Suggestion: might want to use modern Python (https://fstring.help/).

More places like this, won't comment again.

Alternative, more equivalent option: `f"... {count:d}"`.

tests/testenv.py
```diff
@@ -9,3 +11,41 @@ def get_test_result_class(name):
...
+
+    return index, count
+
+
+def shard_list(items, spec):
+    index, count = parse_shard(spec)
+    return items[index - 1::count]
```
robsdedude: Suggested safety net: check whether the sliced items is empty and emit a warning or fail hard to make misconfiguration easier to spot (running an empty shard wastes a fair bit of resources in CI).

robsdedude (approved):
🪓

tests/testenv.py
RichardIrons-neo4j: Good idea, added a hard fail

